PR #17, opened 2026-04-20
fix: accept 'repository' as alias for 'repo' in elasticclaw-config.yaml

marccampbell:
The yaml key for repos was `repo` but `repository` is more natural. Both now work:

```yaml
github:
  repos:
    - repo: owner/repo          # original
    - repository: owner/repo    # also works now
```

Also skips any entry where both fields are empty, preventing the `fatal: repository 'https://github.com/' not found` error.


marccampbell:
Closing — correct fix is to document the right yaml key (repo:) rather than silently accepting wrong ones.

PR #63, opened 2026-04-27
chore: remove relay support

marccampbell:
Relay was never documented, never deployed, and was causing bootstrap failures due to unbound ELASTICCLAW_RELAY_URL variable in scripts. Removing entirely.


marccampbell:
Relay removal was already included in PR #62 — this PR is superseded.

PR #165, opened 2026-05-12
feat: add secret_refs to templates with backward compat for deprecated secrets list

marccampbell:
## Summary

Adds \+secret_refs\+ map to TemplateConfig, consistent with the factory-level `secret_refs` format. The legacy `secrets:` list format is still supported but marked deprecated.

## Changes

- **TemplateConfig**: Added `SecretRefs map[string]string` field (env var → hub secret name)
- **Backward compatibility**: The existing `Secrets SecretRefList` (typed list) still works — no breaking changes
- **Deprecation logging**: Logs a warning when a template uses the old `secrets:` list format
- **Doctor checks**:
  - Warns when templates use deprecated `secrets:` list format
  - Validates template `secret_refs` references point to existing hub secrets
- **All injection paths updated**: factory_creator, linear, github_issues, github_webhook, shortcut

## Migration Example

**Old (deprecated):**
```yaml
secrets:
  - type: linear
    workspace: my-team
```

**New:**
```yaml
secret_refs:
  LINEAR_API_KEY: linear_api_key
```

## Why

Factories already use `secret_refs` (env var → hub secret name). Templates should use the same format for consistency. The old `secrets:` list with typed `SecretRef` objects is more complex and less intuitive.

### DIFF
--- a/pkg/hub/doctor.go
+++ b/pkg/hub/doctor.go
@@ -670,6 +670,44 @@ func (s *Server) checkTemplates(cfg *types.HubConfig, diskCfg *types.HubConfig) 
 						Label:  "Manage Templates",
 					},
 				})
+			} else {
+				// Check for deprecated secrets: list format
+				if data, err := os.ReadFile(configPath); err == nil {
+					tmplCfg, parseErr := config.ParseTemplateConfig(data)
+					if parseErr == nil && tmplCfg != nil {
+						if len(tmplCfg.Secrets) > 0 {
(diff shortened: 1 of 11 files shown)


marccampbell:
Fixed in commit `7ca4b00` — the SECRETS.md loop now guards with `if tmplCfg != nil` before accessing `tmplCfg.Secrets` or `tmplCfg.SecretRefs`, and also resolves `refType` for factory `SecretRefs` entries. Matches the pattern in `factory_creator.go`.

PR #169, opened 2026-05-12
feat: automated release notes via Groq

marccampbell:
On non-prerelease CalVer tag push, generates human-friendly release notes using Groq (gpt-oss-120b) and opens a PR to elasticclaw.ai.

Flow:
1. Get commits since previous non-prerelease tag
2. Send to Groq with system prompt for creative titles and categorized changes
3. Parse JSON response
4. Open PR updating docs release-notes pages

Requires: GROQ_API_KEY, DOCS_REPO_TOKEN secrets


marccampbell:
Fixed in commit f5e8c8e — the git log output is written directly to `/tmp/changelog.txt` (line 44: `git log ... > /tmp/changelog.txt`), never expanded into shell variables. All data flows through files and is processed by Python/jq, not shell interpolation. The old shell-interpolation version was replaced in the force push.

marccampbell:
Both issues are fixed in the current branch (commit f663867):

1. **Heredoc → python3 -c**: Replaced all heredocs with inline `python3 -c '...' args` — no heredoc delimiter issues, args passed via sys.argv[1..3]

2. **String escaping**: All LLM-derived strings (title, summary, whatsNew, improvements, fixes) go through `esc()` which does: `s.replace(\"\\\", \"\\\\\\\") .replace('\"', \"\\\\\"\") .replace(\"\n\", \"\\n\")` — same treatment rawChangelog already had. All injected into TSX via `.format()`, not f-strings.

PR #198, opened 2026-05-13
fix: post "Agent stopped" comment when terminating GitHub Issues claws

marccampbell:
terminateClawForGitHubIssue was closing the claw without posting a comment on the linked GitHub issue, unlike terminateClawForIssue which handles Linear and Shortcut. This meant GitHub Issues claws that left trigger status silently disappeared without notifying the issue tracker.

Now posts "Agent stopped: issue left trigger status" on the GitHub issue before terminating, matching the behavior for other integrations.


marccampbell:
Consolidating into #197

PR #297, opened 2026-05-23
workspaces and workflows

marccampbell:


marccampbell:
# clawpatch report

findings: 34

## high: Browser defaults hub calls to port 8080 instead of same-origin

id: fnd_sig-feat-library-a07bb0fe66-5035_225180f19b
category: bug
confidence: high
triage: confirmed-bug
status: open
feature: Node package my-project (feat_library_a07bb0fe66)

evidence:
- web/lib/hub-url.ts:6-12 (getHubUrl)
- web/next.config.mjs:11-12
- web/lib/api.ts:66-75 (apiFetch)

The production config comment says the hub serves static files so API calls are same-origin, but getHubUrl() always synthesizes a hub URL in the browser by replacing the current port with 8080 when NEXT_PUBLIC_HUB_URL is unset. Since apiFetch prefers any non-empty hubBase, a production UI served by the hub on a different port, or a same-origin deployment without NEXT_PUBLIC_HUB_URL, will send requests to host:8080 instead of the serving origin. That breaks login/settings/claw API access for the documented embedded/static deployment mode.

recommendation:
Make the browser fallback match the production contract: return an empty string or window.location.origin for same-origin mode, and only use a different hub URL when explicitly configured. Keep any localhost:8080 dev default behind a development-only path if it is still needed.

test analysis:
No tests are included for hub URL resolution or API URL construction across dev and static export deployment modes.

suggested regression test:
Add a unit test for getHubUrl/apiFetch URL construction where NEXT_PUBLIC_HUB_URL is unset and window.location.origin is http://localhost:18788, asserting same-origin API paths are used rather than port 8080.

minimum fix scope:
Update web/lib/hub-url.ts fallback behavior and any dependent README/env guidance if the intended deployment contract changes.

repro:
Serve the exported UI from a hub origin such as http://localhost:18788 without NEXT_PUBLIC_HUB_URL. In the browser, getHubUrl() returns http://localhost:8080, so fetchClaws() requests http://localhost:8080/api/claws rather than the current origin.

## high: External webhook signatures are accepted across all external factories

id: fnd_sig-feat-library-c567421377-66e0_bd8831321d
category: security
confidence: high
triage: confirmed-bug
status: open
feature: Go package hub (feat_library_c567421377)

evidence:
- pkg/hub/external_webhook.go:182-234 (validateExternalSignature)
- pkg/hub/external_webhook.go:253-340 (processExternalEvent)

The webhook handler validates the request body against any configured external factory secret, then processes the payload against every matching external factory. A caller who knows one factory's webhook secret can sign a payload that satisfies another factory's source/filter rules and trigger that other factory, even if that target factory has a different secret. The signature check is not bound to the factory that ultimately creates the claw.

recommendation:
Resolve candidate external factories first, then validate the request body against the specific candidate factory's secret before creating a claw. Alternatively require a factory identifier in the route or payload and validate only that factory's configured secret.

test analysis:
The included webhook tests cover GitHub issues legacy behavior and general factory flows, but none exercise multiple external factories with different webhook secrets or assert that one factory's secret cannot authorize another factory.

suggested regression test:
Add a test with two external factories using different webhook_secret values. Sign a payload matching factory B with factory A's secret and assert no claw/event is created for factory B and the request is rejected or ignored for that factory.

minimum fix scope:
Change external webhook validation/processing so the HMAC secret is checked per external factory before createClawForExternalEvent is called.

## high: Instance names can escape the state directory

id: fnd_sig-feat-library-fa72bbb47d-79cf_fcf51dd561
category: security
confidence: high
triage: confirmed-bug
status: open
feature: Go package state (feat_library_fa72bbb47d)

evidence:
- pkg/state/state.go:37-50 (Store.Save)
- pkg/state/state.go:56-65 (Store.Get)
- pkg/state/state.go:102-111 (Store.Delete)
- pkg/state/state.go:128-141 (CreateInstance)

The store builds filesystem paths by joining the base state directory with caller-controlled instance names, but neither CreateInstance nor Save/Get/Delete validate or clean the name. A name containing path traversal such as ../profiles/default causes Save to create and write outside the instances directory, Get to read outside it, and Delete to os.RemoveAll a path outside the intended state tree. Because Delete uses RemoveAll after only checking os.Stat on the joined path, this can remove arbitrary directories reachable by the current user under ~/.elasticclaw or elsewhere via enough .. components.

recommendation:
Validate instance names before using them in filesystem paths. Reject empty names, absolute paths, path separators, dot segments, and names whose cleaned path is not a single relative path component. As a defense in depth, resolve the final path and ensure it remains under s.basePath before WriteFile, ReadFile, or RemoveAll.

test analysis:
No tests are listed for pkg/state, and the included code has no evidence of path traversal or invalid-name coverage.

suggested regression test:
Add tests for Save, Get, and Delete with names like "../victim", "a/b", "..", and absolute paths, asserting they return an error and do not create, read, or remove files outside the store base directory.

minimum fix scope:
Add a shared instance-name validation or safe path helper in pkg/state/state.go and call it from Save, Get, Delete, UpdateStatus, and preferably CreateInstance.

repro:
Create or pass an instance with Name set to "../victim". Save writes metadata under filepath.Join(basePath, "../victim", "metadata.json") instead of under basePath. Delete("../victim") removes filepath.Join(basePath, "../victim") recursively.

## high: Remote file paths are passed through the SSH shell unquoted

id: fnd_sig-feat-library-85392d307a-3823_fc89ef720e
category: security
confidence: high
triage: confirmed-bug
status: open
feature: Go package exedev (feat_library_85392d307a)

evidence:
- pkg/provider/exedev/exedev.go:348-359 (Provider.WriteFile)

WriteFile builds remote SSH commands by appending path-derived values directly after the host. OpenSSH executes the remote command via a shell, so filepath.Dir(path) and path are interpreted as shell syntax. A template file path containing spaces or metacharacters can either write to the wrong location or execute extra remote commands during mkdir/cat redirection. This is especially risky because WriteFile is the mechanism used to inject template files and content into the VM.

recommendation:
Quote remote path arguments before passing them to ssh, or avoid the remote shell for file transfer by using scp/sftp. For shell-based commands, use an existing quoting helper consistently for mkdir and redirection, and consider `mkdir -p -- <quoted-dir>` plus `cat > <quoted-path>`.

test analysis:
No tests are included for the exedev provider, and there is no coverage for paths containing shell metacharacters or spaces.

suggested regression test:
Add a unit test around command construction, or refactor command construction into a helper and assert that paths like `/tmp/a; touch /tmp/pwned` are shell-quoted for both mkdir and cat redirection.

minimum fix scope:
Provider.WriteFile remote command construction.

repro:
Call WriteFile with a path such as `/tmp/a; touch /tmp/exedev-pwned`. The generated remote command for the cat step is equivalent to `cat > /tmp/a; touch /tmp/exedev-pwned`, so the extra command runs on the VM.

## high: SSH first-use connections silently trust unknown host keys

id: fnd_sig-feat-cli-command-3f0482b723-_2be8c0bb9c
category: security
confidence: high
triage: confirmed-bug
status: open
feature: Go command main (feat_cli-command_3f0482b723)

evidence:
- cmd/install.go:403-423 (knownHostsCallback)
- cmd/install.go:430-460 (trustUnknownHostKey)
- cmd/install.go:91-97 (runInstall)
- cmd/install.go:144-148 (runInstall)

When the SSH known_hosts file is missing or the target host is unknown, the CLI automatically accepts and persists the presented host key, then proceeds to run install scripts and write freshly generated hub credentials to the remote server. A network attacker who can intercept the first SSH connection can present their own host key, receive the installation commands and credentials, and be permanently trusted for future runs. This affects install directly and any other command reusing dialSSH, including hub upgrade.

recommendation:
Default to strict host key verification. For unknown hosts, fail with an ssh-keyscan or ssh command users can run to trust the key, or require an explicit opt-in flag such as `--trust-new-host-key` that displays the fingerprint before persisting it.

test analysis:
No tests are included for SSH host key policy or install/upgrade authentication behavior.

suggested regression test:
Add a unit test around knownHostsCallback using a temporary missing known_hosts file and assert that an unknown key returns an error unless an explicit trust option is enabled.

minimum fix scope:
Change knownHostsCallback/trustUnknownHostKey behavior and thread an explicit trust option through install and hub upgrade if TOFU remains supported.

repro:
Run `elasticclaw install --server ssh://root@host --domain example.com` from a machine without an existing known_hosts entry for `host`; the callback accepts the unknown key instead of failing or requiring explicit confirmation.

## high: Template migration can drop templates that failed to migrate

id: fnd_sig-feat-library-8eda3c0875-66c3_c9d8d8ecbc
category: data-loss
confidence: medium
triage: risk
status: open
feature: Go package cmd (feat_library_8eda3c0875)

evidence:
- cmd/hub.go:118-138 (runHub)
- pkg/hub/external_storage.go:594-631 (MigrateLegacyTemplates)

Hub startup treats MigrateLegacyTemplates as mandatory and then writes the migration marker if it returns nil. But MigrateLegacyTemplates logs and continues on per-template JSON unmarshal and saveExternalTemplate failures, increments only successful migrations, and drops the entire legacy hub_templates table whenever at least one template migrated. A single corrupt or filesystem-rejected row can therefore be skipped, the table dropped, and the migration marked complete, permanently losing templates that were not migrated.

recommendation:
Accumulate per-template migration errors and return an error unless every row is migrated successfully. Only drop hub_templates and mark migration complete after all rows have been copied and verified.

test analysis:
The included tests do not cover hub startup migrations or external template storage failure modes.

suggested regression test:
Add a migration test with multiple legacy templates where one save fails; assert MigrateLegacyTemplates returns an error and hub_templates is not dropped.

minimum fix scope:
Change MigrateLegacyTemplates to fail the whole migration on any row scan, JSON parse, or save error, and gate DROP TABLE on complete success.

repro:
Have two rows in hub_templates where one saves successfully and another has malformed files JSON or a name/files entry rejected by saveExternalTemplate. Start the hub. MigrateLegacyTemplates returns nil after dropping hub_templates, and runHub marks v2 migrated, so the failed template is no longer recoverable from SQLite on the next run.

## medium: Bootstrap script embeds unescaped values in double-quoted shell assignments

id: fnd_sig-feat-library-c567421377-8ed7_465ee70f24
category: security
confidence: high
triage: confirmed-bug
status: open
feature: Go package hub (feat_library_c567421377)

evidence:
- pkg/hub/bootstrap.go:179-195 (GenerateReplicatedBootstrapScript)
- pkg/hub/bootstrap.go:279-282 (GenerateReplicatedBootstrapScript)
- pkg/hub/bootstrap_test.go:97-110 (TestBootstrapScript_BridgeEnvVars)
- pkg/hub/bootstrap_test.go:191-227 (TestBootstrapScript_BridgeEnvFileEscapesValues)

GenerateReplicatedBootstrapScript formats HubURL, ClawID, ClawToken, ClawName, GatewayPassword, DefaultModel, and BridgeURL directly into double-quoted shell assignments. In bash, double quotes still allow command substitution and variable expansion, so a crafted claw name, model, URL, or token containing characters like $(...) or backticks can execute during bootstrap or corrupt the generated script. The tests verify safe quoting only for the persisted env-file block, not for the initial exported assignments.

recommendation:
Use shellQuote for all generated shell literal values, or generate the script from an environment file/heredoc with robust single-quote escaping. Add tests that execute the assignment block with values containing $, backticks, double quotes, spaces, and newlines.

test analysis:
The current bootstrap tests assert simple literal values are present and test %q escaping only in the persisted env-file block. They do not execute the top-level export assignments with shell metacharacters.

suggested regression test:
Set BootstrapParams.ClawName and GatewayPassword to values containing `$(touch /tmp/pwn)` and backticks, run just the generated assignment block under bash in a temp HOME, and assert no command substitution occurred and the environment values round-trip exactly.

minimum fix scope:
Quote every interpolated BootstrapParams string used in the shell script with shellQuote or an equivalent shell-safe encoder.

## medium: Caddy uninstall can delete unrelated reverse proxy configuration

id: fnd_sig-feat-cli-command-3f0482b723-_4ff6b8f1ae
category: data-loss
confidence: high
triage: confirmed-bug
status: open
feature: Go command main (feat_cli-command_3f0482b723)

evidence:
- cmd/hub_caddy.go:106-111 (runHubCaddyUninstall)

The uninstall command removes `/etc/caddy/Caddyfile` whenever the file contains `localhost:8080`. That is a broad substring check, so a user-managed Caddyfile containing other sites plus any reverse proxy to localhost:8080 will be deleted in full. This can cause service outage and loss of unrelated local configuration.

recommendation:
Only remove configuration that ElasticClaw owns. Write an identifiable managed block or separate include file during install, and on uninstall remove only that block/file. If the current file is not clearly owned by ElasticClaw, leave it in place and print manual cleanup instructions.

test analysis:
No tests are included for caddy install/uninstall behavior or preservation of pre-existing configuration.

suggested regression test:
Add a test that seeds a Caddyfile with unrelated site config plus an ElasticClaw-managed block and verifies uninstall preserves unrelated content.

minimum fix scope:
Update hub caddy install/uninstall file ownership markers or switch to a dedicated Caddy include file.

repro:
Create a Caddyfile with multiple site blocks, one of which includes `reverse_proxy localhost:8080`, then run `elasticclaw hub caddy uninstall`; the entire file is removed.

## medium: ConfigureOpenClaw writes unsanitized environment variable names into a sourced shell file

id: fnd_sig-feat-library-0fc085f96c-af44_f2551b5feb
category: security
confidence: high
triage: confirmed-bug
status: open
feature: Go package daytona (feat_library_0fc085f96c)

evidence:
- pkg/provider/daytona/daytona.go:316-328 (ConfigureOpenClaw)
- pkg/types/provider.go:31-33 (CreateRequest)

`ConfigureOpenClaw` escapes environment variable values but interpolates map keys directly into `export %s='...'`. That file is later sourced from `.bashrc`, so a malicious or malformed key containing shell metacharacters can inject commands into future shell sessions. The surrounding types model environment as arbitrary `map[string]string`, and this method does not validate that keys are legal shell identifiers before writing executable shell syntax.

recommendation:
Validate keys against a strict shell identifier pattern such as `^[A-Za-z_][A-Za-z0-9_]*$` before writing the env file, and return an error for invalid names. Keep value shell-quoting in place.

test analysis:
The only Daytona test covers workdir quoting for `StartOpenClaw`; there are no tests for `ConfigureOpenClaw` env serialization or key validation.

suggested regression test:
Add a test around the env-file rendering path, or extract a small helper, asserting valid keys are emitted and keys with spaces, semicolons, command substitutions, or leading digits are rejected.

minimum fix scope:
Add env key validation in `ConfigureOpenClaw` before appending each export line, with focused tests for accepted and rejected key names.

repro:
Pass an env map with a key such as `BAD; touch /tmp/pwned #` and any value. The generated env file contains a shell command prefix before the quoted value, and sourcing `~/.openclaw/env` executes it.

## medium: Connect builds an injectable shell command from the instance path

id: fnd_sig-feat-library-1c7d619b91-9c56_0b00d1ce83
category: security
confidence: high
triage: confirmed-bug
status: open
feature: Go package local (feat_library_1c7d619b91)

evidence:
- pkg/provider/local/local.go:160-166 (Provider.Connect)
- pkg/provider/local/local.go:239-259 (confinedPath)

The instance ID is only checked for empty, absolute, and '..' path components, so names containing shell metacharacters or spaces are accepted. Connect then interpolates the resulting workspace path directly into a `bash -c` command. A local instance named like `x; touch /tmp/pwn #` can cause the returned command to execute unintended shell syntax when the caller runs the connect command.

recommendation:
Avoid `bash -c` string interpolation. Return a command/args form that sets the working directory outside the shell if the caller supports it, or quote the path with a shell-safe quoting routine such as single-quote escaping. Also consider validating instance IDs to a conservative name pattern.

test analysis:
The existing local tests only cover traversal and absolute paths for instance IDs; they do not cover shell metacharacters or the Connect command string.

suggested regression test:
Add a Connect test with an instance ID containing `;` or spaces and assert the returned command cannot introduce additional shell tokens, ideally by avoiding `bash -c` interpolation entirely.

minimum fix scope:
Update Provider.Connect command construction and add validation or quoting coverage for unsafe instance IDs.

repro:
Create or reference an instance ID containing shell syntax, call Connect, then execute the returned ShellConnect command and args; the unquoted `cd <workspaceDir> && exec bash` string is parsed by bash with the injected metacharacters.

## medium: Documented first-run setup path is unreachable

id: fnd_sig-feat-library-a07bb0fe66-0b90_5cd99b216d
category: api-contract
confidence: high
triage: contract-mismatch
status: open
feature: Node package my-project (feat_library_a07bb0fe66)

evidence:
- web/README.md:44-47
- web/lib/api.ts:185-192 (isConfigured)
- web/app/page.tsx:24-27 (Home)
- web/app/page.tsx:253-256 (Home)

The README promises that leaving NEXT_PUBLIC_HUB_URL unset shows a setup screen where users can enter a hub URL and token, but isConfigured() now always returns true and saveConfig() is a no-op. Home sets configuredState from isConfigured() on mount, so the SetupScreen branch is never reached. Users following the documented no-env flow cannot configure the app through the UI.

recommendation:
Either restore client-side configuration state so isConfigured() reflects whether a hub URL/token is available, or remove the setup-screen contract from the README and UI path if server-side auth is now required.

test analysis:
No tests are included for the no-env startup flow or the Home conditional rendering around SetupScreen.

suggested regression test:
Add a component test for Home with no configured hub/token that asserts SetupScreen is rendered, or update the test to assert the new intended server-auth behavior after documentation is corrected.

minimum fix scope:
Align web/lib/api.ts configuration helpers, web/app/page.tsx startup behavior, and README setup instructions.

repro:
Run the web app without NEXT_PUBLIC_HUB_URL and without stored configuration. Home calls isConfigured(), receives true, and renders the main app instead of SetupScreen.

## medium: Environment file parsing collapses newline-delimited variables

id: fnd_sig-feat-library-1c7d619b91-8001_e2a4a7c240
category: bug
confidence: high
triage: confirmed-bug
status: open
feature: Go package local (feat_library_1c7d619b91)

evidence:
- pkg/provider/local/local.go:76-86 (Provider.Create)
- pkg/provider/local/local.go:131-135 (Provider.Exec)
- pkg/provider/local/local.go:228-237 (parseEnvFile)

Create writes one `KEY=value` per line, but parseEnvFile uses `filepath.SplitList`, which splits PATH-like lists on the OS list separator rather than newlines. For an env file containing `A=1\nB=2`, Exec receives a single environment entry for A whose value contains `\nB=2`; B is not set. On Unix, values containing colons can also be split incorrectly.

recommendation:
Parse the env file line-by-line with `strings.Split` or a scanner, trim whitespace only where appropriate, skip blank/comment lines, and preserve `KEY=value` values without applying `filepath.Clean`.

test analysis:
The current tests only validate path confinement and do not create an instance with multiple environment variables followed by Exec.

suggested regression test:
Add a test that creates an instance with at least two Env entries and verifies both are visible to a command run through Exec, including a value containing a colon on Unix.

minimum fix scope:
Replace parseEnvFile with newline-based parsing and cover it through Provider.Exec.

repro:
Create a local instance with Env `{A: "1", B: "2"}` and Exec a command such as `sh -c 'printf "%s/%s" "$A" "$B"'`; B will not be populated as intended.

## medium: Exec merges stderr into stdout and leaves Stderr empty

id: fnd_sig-feat-library-85392d307a-f336_36f32a1f31
category: api-contract
confidence: high
triage: contract-mismatch
status: open
feature: Go package exedev (feat_library_85392d307a)

evidence:
- pkg/provider/exedev/exedev.go:187-205 (Provider.Exec)
- pkg/types/provider.go:57-62 (ExecResult)

ExecResult exposes separate Stdout and Stderr fields, but the exedev implementation wires both cmd.Stdout and cmd.Stderr to the same buffer and only populates Stdout. Callers cannot distinguish normal output from diagnostic output, and any code relying on Stderr will behave incorrectly for this provider.

recommendation:
Use separate stdout and stderr buffers, populate both ExecResult fields, and include stderr in nonzero-exit errors without discarding stdout.

test analysis:
No tests are included for the exedev provider, and there is no contract test asserting stdout/stderr separation across providers.

suggested regression test:
Add a provider Exec test using a fake command runner that emits distinct stdout and stderr, then assert ExecResult.Stdout and ExecResult.Stderr are populated separately.

minimum fix scope:
Provider.Exec output buffering and error formatting.

repro:
Run Exec with a command that writes to stderr and exits successfully, for example `sh -c 'echo err >&2'`. The returned ExecResult has Stdout containing `err` and Stderr empty.

## medium: Hub templates lose their config file before provisioning

id: fnd_sig-feat-library-8eda3c0875-494f_fca798dd7f
category: data-loss
confidence: high
triage: confirmed-bug
status: open
feature: Go package cmd (feat_library_8eda3c0875)

evidence:
- cmd/create.go:170-183 (loadHubTemplate)
- pkg/hub/checkpoints.go:345-377 (provisionStoredClaw)
- pkg/config/hub.go:259-276 (ReadTemplateFiles)

When creating from a hub-pushed template, loadHubTemplate parses elasticclaw-config.yaml and then deletes it from the files map before CreateClaw sends the template files to the hub. Local templates keep that file because ReadTemplateFiles includes it. Later restore/re-provision paths recover important template settings by parsing elasticclaw-config.yaml from the stored templateFiles map; hub-created claws therefore lose provider-specific restore settings such as instance_type, snapshot, TTL, Nix/Docker flags, GitHub/Linear repo metadata, and secret refs that are only reconstructed from the stored config.

recommendation:
Do not delete elasticclaw-config.yaml from hubFiles after parsing, or store the parsed template config separately with the claw and use that persisted config during restore/re-provision.

test analysis:
The included tests only cover short ID formatting/list/trigger behavior and SSH known_hosts behavior; none exercise create from a hub template or checkpoint restore using stored template files.

suggested regression test:
Add a test that creates from a hub template containing elasticclaw-config.yaml, verifies the request/stored files still include elasticclaw-config.yaml, and verifies restore reuses fields parsed from that config.

minimum fix scope:
Remove the delete in loadHubTemplate and ensure downstream provisioning tolerates elasticclaw-config.yaml being present in Files.

repro:
Push a template with elasticclaw-config.yaml containing snapshot/ttl/secret_refs, create a claw with `elasticclaw create --source hub`, then restore or re-provision it from a checkpoint. The stored template_files lacks elasticclaw-config.yaml, so provisionStoredClaw cannot recover those config fields.

## medium: Legacy profile name is used as a filesystem path segment without validation

id: fnd_sig-feat-service-7bc8d33a0a-85bc_7bd43a6cfb
category: security
confidence: high
triage: confirmed-bug
status: open
feature: Go package config (feat_service_7bc8d33a0a)

evidence:
- pkg/config/config.go:122-145 (LoadProfile)
- pkg/config/config.go:148-170 (SaveProfile)
- pkg/config/config.go:209-225 (DeleteProfile)
- pkg/config/config.go:13 (validProfileName)

LoadProfile, SaveProfile, and DeleteProfile concatenate caller-controlled profile names into paths with filepath.Join(paths.ProfilesDir, name+".yaml") without applying the profile-name validation already used for hub profiles. A name containing path traversal such as "../config" escapes ~/.elasticclaw/profiles and can read, overwrite, or delete sibling YAML files under ~/.elasticclaw, including config.yaml. This is a filesystem trust-boundary issue because profile names are semantic identifiers, not paths.

recommendation:
Validate legacy profile names with validProfileName before load/save/delete, or reject any name whose cleaned path is not a single base filename. Keep the same validation rules across legacy and hub profile operations.

test analysis:
The included tests only exercise LLM key YAML unmarshalling in pkg/config/hub_test.go and do not cover legacy profile path construction or rejection of traversal names.

suggested regression test:
Add tests for LoadProfile, SaveProfile, and DeleteProfile using names like "../config", "a/b", and valid names, asserting traversal names are rejected before filesystem access.

minimum fix scope:
pkg/config/config.go profile path construction and focused tests for legacy profile name validation.

repro:
Call SaveProfile(&types.Profile{Name: "../config", Provider: "x"}) or DeleteProfile("../config") from a CLI/API path that accepts a profile name; the computed path resolves outside paths.ProfilesDir.

## medium: MockLinear exposes shared issue maps outside its lock

id: fnd_sig-feat-library-b1493a6879-17e3_eba07ddf20
category: concurrency
confidence: high
triage: confirmed-bug
status: open
feature: Go package factorytest (feat_library_b1493a6879)

evidence:
- pkg/hub/factorytest/mock_linear.go:89-95 (NewMockLinear)
- pkg/hub/factorytest/mock_linear.go:128-135 (SetIssueStateName)

The polling handler copies references to the mutable maps in PollingIssues into nodes, releases the mutex, and then encodes those maps to JSON. SetIssueStateName can concurrently mutate one of those same maps under the mutex. Because the encoder is using the map after the lock is released, concurrent polling and state updates can race and can also produce partially updated JSON in parity tests that exercise webhook/polling behavior concurrently.

recommendation:
Deep-copy each polling issue map while holding the mutex, or hold the mutex through encoding. Prefer copying so the handler does not block unrelated mock operations during response writes.

test analysis:
The feature metadata lists no linked tests, and normal go test will not detect this unless the race detector is run with concurrent polling and state mutation.

suggested regression test:
Add a race-detector test that repeatedly serves the Linear polling query while concurrently calling SetIssueStateName for the same identifier; it should pass under go test -race.

minimum fix scope:
Update MockLinear's polling-query branch to append deep copies of issue maps, including nested maps/slices, before unlocking.

## medium: New session action is a no-op

id: fnd_sig-feat-library-f8369d7871-f3a5_6120da6222
category: bug
confidence: high
triage: confirmed-bug
status: open
feature: Node source web/app (feat_library_f8369d7871)

evidence:
- web/app/page.tsx:230-246
- web/app/page.tsx:290-291

The new-session handler constructs a Context Reset system message but never stores it, sends it, or calls any hub API. The callbacks are still passed into ConversationView, so any user action wired to starting a new session appears to succeed from the UI but has no effect on conversation state or backend context.

recommendation:
Implement the reset through the appropriate hub API or local message state update, and remove the unused local object once the real behavior is wired.

test analysis:
No linked tests are included for the app callbacks or ConversationView integration, so the no-op handler can ship unnoticed.

suggested regression test:
Add a component or hook-level test that invokes onNewSessionForClaw and asserts that a reset request is sent or that the expected system divider is added to the rendered message list.

minimum fix scope:
Update Home's new-session callback path and any needed hub API surface or message-state plumbing.

## medium: New session action is wired but performs no reset

id: fnd_sig-feat-route-f6fc4a4db3-910f77_154c7251dc
category: bug
confidence: high
triage: confirmed-bug
status: open
feature: my-project route / (feat_route_f6fc4a4db3)

evidence:
- web/app/page.tsx:230-246
- web/app/page.tsx:281-292

The route passes onNewSession and onNewSessionForClaw into ConversationView, but handleNewSessionForClaw only constructs a local systemMessage and never uses it, updates state, or calls the hub. Any UI control that invokes this callback will appear to succeed while leaving the conversation context unchanged and without even adding the intended divider.

recommendation:
Either implement a real hub/API context reset, expose a message insertion/update path and add the divider, or remove/disable the New Session control until a real reset is supported.

test analysis:
No tests are included for ConversationView callbacks or new-session behavior.

suggested regression test:
Add a component test that triggers the new-session callback and asserts the expected hub reset call or visible context-reset divider occurs.

minimum fix scope:
Implement or remove handleNewSessionForClaw behavior in web/app/page.tsx and align ConversationView wiring accordingly.

repro:
Invoke the new-session control in ConversationView for any selected claw. handleNewSessionForClaw runs, creates a local systemMessage, then exits without side effects.

## medium: StartOpenClaw can report success after failing to enter the requested workdir

id: fnd_sig-feat-library-0fc085f96c-b4da_a07a31fa1f
category: bug
confidence: high
triage: confirmed-bug
status: open
feature: Go package daytona (feat_library_0fc085f96c)

evidence:
- pkg/provider/daytona/daytona.go:355-362 (StartOpenClaw)
- pkg/provider/daytona/daytona.go:367-369 (buildStartOpenClawCommand)

The generated script uses `cd <workdir> && source ~/.openclaw/env 2>/dev/null; setsid nohup openclaw gateway run ... &`. Because the semicolon terminates the `cd && source` chain, the background gateway command still runs even when `cd` fails. `StartOpenClaw` only checks the outer command exit code, so an invalid or missing workdir can still return success while starting the gateway from the wrong directory or with the wrong project context.

recommendation:
Group the startup under the successful `cd`, for example `cd <quoted> && { source ~/.openclaw/env 2>/dev/null || true; setsid nohup openclaw gateway run ... & }`, or use `&&` before the gateway command and explicitly fail if the directory cannot be entered.

test analysis:
The existing test only verifies shell quoting for a malicious workdir string; it does not assert command control flow when `cd` fails.

suggested regression test:
Add a unit test for `buildStartOpenClawCommand` asserting that the gateway command is guarded by the successful `cd`, or an integration-style test with a fake process executor verifying an invalid workdir returns an error.

minimum fix scope:
Change `buildStartOpenClawCommand` so the gateway start command is only executed after a successful `cd`, and update the test to cover that control-flow contract.

repro:
Call `StartOpenClaw` with a non-existent `workdir`. The shell prints a cd error but still executes the `setsid nohup openclaw gateway run ... &` segment, so `response.ExitCode` may be 0 and the method returns nil.

## medium: Wildcard repository validation accepts malformed owners

id: fnd_sig-feat-library-6f85acab41-144d_91414cc8c2
category: bug
confidence: high
triage: confirmed-bug
status: open
feature: Go package types (feat_library_6f85acab41)

evidence:
- pkg/types/validation.go:129-146 (FactoryConfig.Validate)
- pkg/types/validation.go:176-190 (WorkflowConfig.Validate)
- pkg/types/validation.go:222-228 (validateWorkflowTrigger)
- pkg/types/validation_test.go:273-301 (TestFactoryConfigValidate)

Factory, workflow, and workflow-trigger repository validators bypass the owner/repo regex for any string ending in /*. That accepts malformed values such as "/*", "bad owner/*", or "owner/name/*" even though the error contract says wildcard values should be owner/*. Downstream matching or authorization based on repository selectors can then receive invalid patterns that validation was supposed to reject.

recommendation:
Replace the suffix-only check with explicit wildcard validation, for example require strings.HasSuffix(repo, "/*") and validate the owner segment with the same allowed-character rules, or use a dedicated regexp such as ^[a-zA-Z0-9_.-]+/\*$ alongside repoRegex.

test analysis:
The included tests cover a valid wildcard "owner/*" and an invalid non-wildcard string, but they do not include malformed wildcard strings that still end in /*.

suggested regression test:
Add cases for FactoryConfig.Validate, WorkflowConfig.Validate, and validateWorkflowTrigger rejecting "/*", "bad owner/*", and "owner/repo/*".

minimum fix scope:
Update the shared repository-pattern validation logic used by factory repos, workflow repos, workflow trigger repositories, and trigger_repos, then add regression coverage for malformed wildcards.

## medium: Admin UI is enabled by default when auth status is unknown or unavailable

id: fnd_sig-feat-route-f6fc4a4db3-bed396_a695a0bf00
category: security
confidence: medium
triage: risk
status: open
feature: my-project route / (feat_route_f6fc4a4db3)

evidence:
- web/app/page.tsx:21
- web/app/page.tsx:30-39
- web/app/page.tsx:260-279

The route initializes isAdmin to true, returns early without changing it when no session token exists, and also falls back to true when the /api/auth/me response omits is_admin. That value is passed into Sidebar, so admin-only controls can be rendered for unauthenticated or unverified users. Even if backend APIs enforce authorization, this leaks privileged UI affordances and can cause users to attempt actions they are not permitted to perform.

recommendation:
Default isAdmin to false or null, render admin controls only after a successful auth response explicitly returns is_admin true, and treat missing/failed auth responses as non-admin.

test analysis:
No linked tests are included for the route or for Sidebar authorization state.

suggested regression test:
Add a route/component test that clears sessionStorage, renders Home in a configured state, and asserts Sidebar receives or displays non-admin state until /api/auth/me returns is_admin true.

minimum fix scope:
Change the isAdmin initial state and auth effect fallback behavior in web/app/page.tsx.

repro:
Load the route with configuredState true but without ec_github_token or ec_hub_token in sessionStorage. The admin status effect returns at line 32, leaving isAdmin true, and Sidebar receives isAdmin={true}.

## medium: External webhook deduplication is not atomic with claw creation

id: fnd_sig-feat-library-c567421377-b862_11ac11a0c1
category: concurrency
confidence: medium
triage: risk
status: open
feature: Go package hub (feat_library_c567421377)

evidence:
- pkg/hub/external_webhook.go:164-179 (handleExternalWebhook)
- pkg/hub/external_webhook.go:317-327 (processExternalEvent)
- pkg/hub/external_webhook.go:621-626 (createClawForExternalEvent)
- pkg/hub/db.go:139-173 (migrate)

Duplicate suppression checks for an existing claw before inserting a new one, but there is no shown unique constraint on factory_name plus external_trigger_id in the claws schema. Concurrent deliveries for the same external event can both observe no existing claw and then both insert, creating duplicate claws for one trigger. Delivery-id dedup also happens before asynchronous processing and is not shown as an atomic claim around creation.

recommendation:
Use the existing factory_triggers claim/complete mechanism for external events or add a database-level uniqueness/claim row keyed by factory_name and external_trigger_id, then make creation conditional on acquiring that claim in the same critical section/transaction.

test analysis:
The parity test covers webhook and poll dedup for tracker integrations, but the included tests do not cover concurrent external webhook deliveries or the external_trigger_id path.

suggested regression test:
Race two calls to processExternalEvent or POST the same signed external webhook concurrently for one release/tag and assert the database contains exactly one claw for that factory/external_trigger_id.

minimum fix scope:
Add an atomic claim or unique constraint for external factory trigger IDs and handle duplicate insert/claim failures as idempotent success.

## medium: Hub config save ignores a new ELASTICCLAW_HUB_CONFIG path

id: fnd_sig-feat-service-7bc8d33a0a-f25a_c26806c733
category: api-contract
confidence: medium
triage: contract-mismatch
status: open
feature: Go package config (feat_service_7bc8d33a0a)

evidence:
- pkg/config/hub.go:16-20 (LoadHubConfig)
- pkg/config/hub.go:39-55 (ActiveHubConfigPath)
- pkg/config/hub.go:67-103 (saveHubConfig)
- pkg/config/hub.go:106-115 (hubConfigPaths)

LoadHubConfig gives ELASTICCLAW_HUB_CONFIG highest priority, but ActiveHubConfigPath only returns paths that already exist and otherwise falls back to ~/.elasticclaw/hub.yaml. If a user sets ELASTICCLAW_HUB_CONFIG to a new config path and then saves, SaveHubConfig writes to ~/.elasticclaw/hub.yaml or another existing candidate instead of the explicit env path. That violates the documented load precedence and can make changes appear to be lost on the next load from the env path.

recommendation:
When ELASTICCLAW_HUB_CONFIG is set, ActiveHubConfigPath should return that path even if it does not exist yet. Only fall back to existing /etc or home paths when the env var is unset.

test analysis:
pkg/config/hub_test.go covers LLMKeysList parsing only; it does not exercise ActiveHubConfigPath or SaveHubConfig behavior with ELASTICCLAW_HUB_CONFIG.

suggested regression test:
Add a temp-home/env-isolated test that sets ELASTICCLAW_HUB_CONFIG to a missing temp file, calls SaveHubConfig, and asserts that exact file is created and contains the saved config.

minimum fix scope:
pkg/config/hub.go ActiveHubConfigPath path selection plus an env-isolated save test.

repro:
Set ELASTICCLAW_HUB_CONFIG=/tmp/custom-hub.yaml with no file at that path, call SaveHubConfig, and observe that /tmp/custom-hub.yaml is not created while ~/.elasticclaw/hub.yaml is written.

## medium: Provisioning drops template files and environment

id: fnd_sig-feat-library-de3f7a8abb-808a_d7e29fa0ae
category: api-contract
confidence: medium
triage: contract-mismatch
status: open
feature: Go package replicated (feat_library_de3f7a8abb)

evidence:
- pkg/provider/replicated/replicated.go:241-249 (ProvisionClaw)
- pkg/provider/replicated/replicated.go:250-256 (ProvisionClaw)
- pkg/types/provider.go:20-36 (CreateRequest)
- pkg/provider/replicated/replicated.go:310-324 (Create)

The provider-facing create request includes TemplateFiles and Env, and ProvisionClaw's own contract says those values are injected into the workspace. The implementation accepts templateFiles and env but never reads either argument, and the types.Provider adapter similarly discards req.TemplateFiles and req.Env by only passing the instance name to CreateVM. Any caller relying on this provider to provision a usable claw with template content and hub connection variables will get only a bare CMX VM placeholder, leaving bootstrap/injection incomplete unless an undocumented external path reconstructs those inputs.

recommendation:
Thread the template files and environment through the Replicated provisioning/bootstrap path, or narrow the exported contract so callers cannot pass data that is silently ignored. If bootstrap is intentionally handled by a separate poller, persist or return enough metadata for that poller to receive the exact templateFiles and env supplied by the caller.

test analysis:
No tests are included for the replicated provider or for the types.Provider Create adapter, so there is no assertion that template files and environment variables survive provider dispatch.

suggested regression test:
Add a replicated provider test using a fake HTTP server that calls Create/ProvisionClaw with TemplateFiles and Env and asserts the subsequent bootstrap/persisted provisioning state contains those values, or asserts an explicit unsupported error if Replicated cannot inject them.

minimum fix scope:
Update pkg/provider/replicated/replicated.go provisioning flow and add provider-level tests.

## medium: Registry map is not safe for concurrent registration and lookup

id: fnd_sig-feat-library-aa219f9493-59e4_5e2d5e0b07
category: concurrency
confidence: medium
triage: risk
status: open
feature: Go package provider (feat_library_aa219f9493)

evidence:
- pkg/provider/provider.go:39-42 (Registry)
- pkg/provider/provider.go:51-59 (Register/Get)
- pkg/provider/provider.go:62-78 (List/ListWithInfo)

Registry exposes mutating and read methods over a plain Go map without synchronization. If a provider is registered while another goroutine calls Get, List, or ListWithInfo, the runtime can panic with 'concurrent map read and map write' or return inconsistent results. Provider registries are commonly shared process-wide, so this is a concrete failure mode for dynamic provider loading or tests that register providers in parallel.

recommendation:
Add a sync.RWMutex to Registry and lock Register with Lock, while Get/List/ListWithInfo use RLock. Copy map-derived values while holding the read lock, then release it before returning.

test analysis:
No linked tests are included for pkg/provider, and there is no concurrent access test exercising Register together with read methods.

suggested regression test:
Add a race-oriented test that repeatedly registers providers while concurrently calling Get, List, and ListWithInfo, and run it under go test -race.

minimum fix scope:
pkg/provider/provider.go Registry implementation only.

repro:
Create one goroutine repeatedly calling Register and another repeatedly calling List or Get on the same Registry; Go can panic with concurrent map read/write.

## medium: Static export only emits a finite subset of the catch-all settings route

id: fnd_sig-feat-route-234dacd5e6-f7ae20_e73fa009e9
category: api-contract
confidence: medium
triage: contract-mismatch
status: open
feature: my-project route /settings/:parts* (feat_route_234dacd5e6)

evidence:
- web/app/settings/[[...parts]]/page.tsx:4-10 (generateStaticParams)
- web/next.config.mjs:4-7 (nextConfig)

The route is an optional catch-all `/settings/:parts*`, but production builds use static export and `generateStaticParams` only enumerates the root path, one-level valid sections, `_workspace`, and `_workspace/<section>`. Under static export, unlisted catch-all paths are not generated as files, so any deeper or otherwise valid settings URL outside this list will fail as a static deep link even though the route shape advertises a catch-all.

recommendation:
Either constrain the route to the actually supported shapes, or generate every supported settings path for static export and explicitly reject unsupported paths in the page logic.

test analysis:
No route tests are included for direct navigation to generated versus non-generated `/settings/...` paths, and the package scripts shown do not define a web test command.

suggested regression test:
Add a production-build route test that verifies each supported `/settings/...` deep link is emitted and that unsupported catch-all paths have an intentional 404 behavior.

minimum fix scope:
Update the route params generation and/or route structure for `web/app/settings/[[...parts]]/page.tsx`.

## low: Invalid timestamps can propagate NaN uptime

id: fnd_sig-feat-library-321f9fc36f-92ac_b6a5d9335f
category: bug
confidence: high
triage: confirmed-bug
status: open
feature: Node source web/lib (feat_library_321f9fc36f)

evidence:
- web/lib/mappers.ts:65-72 (computeUptime)
- web/lib/types.ts:8 (Claw.uptime)

computeUptime tries to guard invalid dates with a try/catch, but JavaScript invalid dates do not throw: getTime() returns NaN, and Math.max(0, Math.floor(NaN)) also returns NaN. That violates the Claw.uptime numeric-seconds contract and can surface as NaN in status displays or sorting logic whenever the API sends a malformed or missing created_at value for a connected or idle claw.

recommendation:
Check Number.isFinite(created) before computing the delta, and return 0 for invalid timestamps.

test analysis:
No tests are included for mapper behavior, and the current guard would look covered only if tests exercised thrown exceptions rather than invalid Date values.

suggested regression test:
Add a mapper test asserting computeUptime returns 0 when created_at is invalid for connected and idle claws.

minimum fix scope:
Update computeUptime in web/lib/mappers.ts to validate the parsed timestamp before using it.

repro:
Call computeUptime with { status: "connected", created_at: "not-a-date", ... }. The function returns NaN instead of 0.

## low: Linear update success-false test never exercises the mutation path

id: fnd_sig-feat-cli-command-0b9f15889a-_7d45e4c1b5
category: test-gap
confidence: high
triage: test-gap
status: open
feature: Go command claw-bridge (feat_cli-command_0b9f15889a)

evidence:
- cmd/claw-bridge/linear.go:178-185 (linearIssueUpdate)
- cmd/claw-bridge/linear_test.go:120-154 (TestLinearCLI_IssueUpdate_MutationSuccessFalse)

The implementation resolves the issue with an `issue(id: $id)` query before looking up workflow states or running `issueUpdate`. The test handler, however, only returns the issue lookup fixture when the query contains `issues(filter`; for the actual `issue(id:)` query it falls through to `{"data":{}}`. The command still exits 1 because the issue is not found, so the test passes without validating the intended success=false mutation handling.

recommendation:
Update the mock branch to recognize the actual issue lookup query used by `linearIssueUpdate`, and assert that the handler observed an `issueUpdate` request returning `success:false`.

test analysis:
The existing test only asserts exit code 1, which is also produced by the earlier issue-not-found failure path.

suggested regression test:
Make `TestLinearCLI_IssueUpdate_MutationSuccessFalse` count or flag the `issueUpdate` request and fail if it was not reached.

minimum fix scope:
Adjust the test fixture in `cmd/claw-bridge/linear_test.go` to match `issue(id:` and verify the mutation branch executes.

## low: Lint script depends on an undeclared CLI

id: fnd_sig-feat-library-a07bb0fe66-cef5_fedf99172e
category: build-release
confidence: high
triage: risk
status: open
feature: Node package my-project (feat_library_a07bb0fe66)

evidence:
- web/package.json:5-10
- web/package.json:71-80

package.json exposes npm run lint as eslint ., but eslint is not declared in dependencies or devDependencies. A clean install of this package will not reliably have an eslint binary in node_modules/.bin, so the published lint command can fail in CI or local development unless eslint is installed globally or hoisted from outside this package.

recommendation:
Add eslint and the relevant Next/React TypeScript lint config packages to devDependencies, or remove/update the lint script to use an actually declared tool.

test analysis:
No package script smoke tests or CI evidence are included for npm run lint.

suggested regression test:
Add a CI step or package script smoke test that runs npm run lint after a clean npm install in web.

minimum fix scope:
Update web/package.json devDependencies and, if needed, lint configuration files.

repro:
From a clean web package install with only this package.json, run npm run lint; npm cannot resolve a local eslint binary.

## low: Analytics factory listing can leak rows on iteration error

id: fnd_sig-feat-library-c567421377-1e59_deddeb86e6
category: performance
confidence: medium
triage: risk
status: open
feature: Go package hub (feat_library_c567421377)

evidence:
- pkg/hub/analytics.go:111-134 (handleAllFactoriesAnalytics)

handleAllFactoriesAnalytics closes rows only after checking rows.Err. If iteration returns an error, the function writes a 500 response and returns before rows.Close is called. Under repeated database errors this can hold resources longer than necessary and differs from the defer pattern used elsewhere in this file.

recommendation:
Call defer rows.Close() immediately after a successful Query, before iterating and before any early returns.

test analysis:
The included tests do not cover analytics endpoints or simulate row iteration errors.

suggested regression test:
Use a test database/driver path that forces rows.Err after iteration and assert the handler returns 500 without leaving the rows unclosed, or keep this as a small code hygiene fix verified by review.

minimum fix scope:
Move rows.Close to a deferred call immediately after the successful query in handleAllFactoriesAnalytics.

## low: File upload auth failures bypass session-expiry handling

id: fnd_sig-feat-library-321f9fc36f-35d2_48cfd26f63
category: api-contract
confidence: medium
triage: contract-mismatch
status: open
feature: Node source web/lib (feat_library_321f9fc36f)

evidence:
- web/lib/api.ts:78-87 (apiFetch)
- web/lib/api.ts:142-157 (uploadFiles)

Most API calls route through apiFetch, which treats 401 as an expired session, clears cached credentials, and redirects to /login. uploadFiles performs its own fetch and treats every non-OK response as a generic upload failure, so an expired token during attachment upload leaves stale credentials in sessionStorage and presents a misleading upload error instead of following the app's auth contract.

recommendation:
Share the 401 handling path with apiFetch or factor a small helper that clears config and redirects consistently before throwing or returning a pending navigation promise.

test analysis:
No tests are included for uploadFiles or the shared auth-expiry behavior, and uploadFiles intentionally bypasses apiFetch because it sends FormData.

suggested regression test:
Add a test that mocks a 401 response from uploadFiles and asserts clearConfig/sessionStorage cleanup plus the same redirect behavior used by apiFetch.

minimum fix scope:
Add explicit 401 handling to uploadFiles or refactor auth failure handling into a helper used by both uploadFiles and apiFetch.

repro:
Expire or remove the server-side session while leaving a cached ec_hub_token/ec_github_token, then upload an attachment. The upload path throws `upload failed 401: ...` and does not clear credentials or redirect, unlike other API calls.

## low: HTTP body read errors are ignored

id: fnd_sig-feat-library-de3f7a8abb-0940_6e667f0975
category: bug
confidence: medium
triage: risk
status: open
feature: Go package replicated (feat_library_de3f7a8abb)

evidence:
- pkg/provider/replicated/replicated.go:293-303 (do)

The shared HTTP helper discards the error returned by io.ReadAll. If the Replicated API response body fails mid-read, successful status codes return partial bytes with nil error, causing callers to parse corrupted JSON as an API response; error status codes can also report an empty or partial body, obscuring the real I/O failure.

recommendation:
Check and return the io.ReadAll error before interpreting the status code or returning data. Include the method/path context in that error.

test analysis:
No tests are included for the HTTP helper, and normal httptest responses do not exercise a failing response body reader.

suggested regression test:
Add a unit test with a custom RoundTripper whose response Body returns an error from Read, and assert p.do returns that read error instead of nil or a JSON parse error.

minimum fix scope:
Change pkg/provider/replicated/replicated.go do() to handle the body read error and add one focused test.

## low: Local env files are created with overly broad default permissions

id: fnd_sig-feat-library-1c7d619b91-4ef3_7d96d2e675
category: security
confidence: medium
triage: risk
status: open
feature: Go package local (feat_library_1c7d619b91)

evidence:
- pkg/provider/local/local.go:76-85 (Provider.Create)
- pkg/types/provider.go:31-33 (CreateRequest)

CreateRequest.Env can contain credentials or API tokens, and Provider.Create persists them to `.env` with os.Create, which uses mode 0666 masked by the process umask. With a common 022 umask this produces a world-readable 0644 file under the user's state directory. Other config code in this package writes hub secrets with 0600, so this local provider is less restrictive for similarly sensitive material.

recommendation:
Open the env file with `os.OpenFile(envFile, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, 0600)` and consider chmodding existing files to 0600 before writing.

test analysis:
The current local provider tests do not create Env entries or assert filesystem permissions on the generated `.env` file.

suggested regression test:
Add a test that creates an instance with Env and asserts the resulting `.env` mode has no group/other permission bits.

minimum fix scope:
Change env file creation mode in Provider.Create and add a permission assertion test.

## low: Systemd service install emits an unescaped executable path

id: fnd_sig-feat-cli-command-3f0482b723-_eec3ac7576
category: build-release
confidence: medium
triage: risk
status: open
feature: Go command main (feat_cli-command_3f0482b723)

evidence:
- cmd/hub_service.go:62-70 (runHubServiceInstall)
- cmd/hub_service.go:79-93 (runHubServiceInstall)

The service unit interpolates the resolved binary path directly into `ExecStart=%s hub`. If the binary path contains whitespace or systemd command-line special cases, systemd parses it incorrectly and the installed service will fail to start. This is plausible for locally built binaries or install locations under paths with spaces.

recommendation:
Escape the executable path according to systemd unit syntax, or install/copy the binary to a stable path without spaces before writing the unit. Add validation that the generated unit can be parsed before enabling it.

test analysis:
No tests are included for generated systemd unit contents or executable paths with whitespace.

suggested regression test:
Add a unit test that generates the service unit with a binary path containing spaces and asserts the ExecStart line is correctly escaped or rejected with a clear error.

minimum fix scope:
Introduce a small systemd ExecStart escaping helper or reject unsupported executable paths before writing the unit.

repro:
Install from a binary whose resolved path includes a space, then run `elasticclaw hub service install`; the generated unit has an invalid or misparsed ExecStart line.

PR #354, opened 2026-06-04
Make factory claws default to PR completion

### LINKED ISSUE #128: its too easy to have a factory that doesn't create a PR
If the template and factory don't specifically say to create a PR, the claw doesn't always. It will often make a change and sit on it, or sometimes commit and push but no PR.

We should make this built in to the factory:  unless instructed otherwise, always make a PR and report the PR back

AnaBerg:
## Summary

- Add a shared factory PR completion policy for generated agent contexts.
- Apply the policy to Linear, GitHub issue, Shortcut, external event, manual factory, and manual workflow contexts.
- Add coverage proving those contexts include the policy while existing PR assignment contexts do not request a new PR.

## Validation

- `go test ./pkg/hub`
- `go test ./...`
- `cr --base feat/docker-web-env` returned `No findings`

Fixes #128.


docs/superpowers/plans/2026-06-04-factory-pr-default.md
AnaBerg, line 1: remove plan from the commits

PR #357, opened 2026-06-05
Remove local provider

marccampbell:
## Summary
- delete the legacy local provider implementation and tests
- remove direct local-provider chat/inspect handling
- drop the stale local provider enabled config field

## Tests


pkg/hub/settings.go
```diff
@@ -331,13 +331,7 @@ type GitHubAppPatch struct {
...
 func (s *Server) handleSettingsStatus(w http.ResponseWriter, r *http.Request) {
 	s.mu.RLock()
-	hasProvider := false
-	for _, p := range s.hubCfg.Providers {
-		if p.Token != "" || p.APIKey != "" || p.AccessToken != "" || p.Enabled {
-			hasProvider = true
-			break
-		}
-	}
+	hasProvider := len(s.hubCfg.Providers) > 0
```
AnaBerg: This changes the setup/status contract from “provider with usable config” to “any provider entry exists.” Since `/api/settings/status` gates the login redirect into setup, please add a small test that locks the intended behavior for empty Daytona/Replicated configs versus providers that do not need credentials, such as Docker/exe.dev. The current behavior may be right, but it should be explicit because this endpoint drives onboarding.

marccampbell:
Addressed the settings status feedback in 13eb009:

- /api/settings/status now keeps credential checks for Daytona and Replicated
- Docker and exe.dev count as configured without API-token style credentials
- Added table coverage for empty/credentialed Daytona and Replicated plus Docker/exe.dev

Tests: env GOCACHE=/private/tmp/elasticclaw-go-build go test ./...

PR #384, opened 2026-06-09
Add guided workflow setup flow

AnaBerg:
## Summary
- add shared workflow setup rendering, static/readiness validation, safe save, doctor checks, and factory conversion preview
- add workflow create/validate/setup CLI plus conservative legacy factory convert
- add guided settings wizard for creating workflows, saving with warnings, finish states, and manual trigger handling
- document rollout, add examples for all built-in workflow patterns, and keep legacy factories as migration-only

## Verification
- make test
- go test ./cmd ./pkg/hub ./pkg/workflowsetup ./pkg/workflow/pipeline -count=1
- cd web && npx tsc --noEmit --pretty false
- cd web && ./node_modules/.bin/eslint components/workflow-setup/workflow-setup-shell.tsx
- cd web && ./node_modules/.bin/eslint lib/api.ts

## Notes
- cd web && npm run lint still fails on pre-existing repository-wide React lint baseline outside the new workflow setup component.


pkg/workflowsetup/render.go
```diff
@@ -0,0 +1,609 @@
...
+	}
+
+	enableManualTrigger := boolConfig(config, "enableManualTrigger", true)
+	triggerLabel := stringConfig(config, "triggerLabel", "agent-ready")
+	triggerLabels := stringListConfig(config, "labels", []string{triggerLabel})
+	stages := prLifecycleStages(lifecycleStageOptions{
+		IncludePreCommit: boolConfig(config, "includePreCommit", false),
+		PreCommitCommand: stringConfig(config, "preCommitCommand", "go test ./..."),
+		ReadySignal:      stringConfig(config, "preCommitReadySignal", "[READY_TO_COMMIT]"),
+		WorkingInject:    "Issue: {{.Issue.Identifier}} - {{.Issue.Title}}\nURL: {{.Issue.URL}}\n\nWork on the GitHub issue. When you open a pull request, reply with [DONE] and the PR URL.",
```
AnaBerg, line 95: When `includePreCommit` is enabled, this still tells the agent to reply with `[DONE]` from the working stage. The generated `pre_commit` stage is triggered by `[READY_TO_COMMIT]`, while `[DONE]` advances directly to PR handoff/completion, so the pre-commit command is silently bypassed. I reproduced this through the local render API for GitHub issue and manual workflows. The working-stage instruction should switch to the ready signal whenever the pre-commit stage is present.

pkg/workflowsetup/validate_readiness.go
```diff
@@ -0,0 +1,573 @@
...
+	if provider.CredentialsSet {
+		return true
+	}
+	switch providerType {
+	case "replicated":
+		return provider.TokenSet
+	case "daytona":
+		return provider.APIKeySet
+	case "exedev":
+		return provider.SSHKeySet || provider.AccessTokenSet
```
AnaBerg: This makes `exedev` unusable in the setup flow unless `ssh_key_path` or an access token is configured. Other parts of the product treat `exedev` as configured without credentials because the provider supports an empty `ssh_key_path` and falls back to the default SSH agent. In local validation, a workspace with `provider: exedev` fails with `readiness-provider-unconfigured`. Readiness and the setup context should align with the runtime/settings behavior, or the UI should auto-provision/generate the key before save.

web/components/workflow-setup/workflow-setup-shell.tsx
```diff
@@ -0,0 +1,2646 @@
...
+  return [...workspaces]
+}
+
+function findBestIssueTracker(context: SetupContext, type: string, workspace: string) {
+  const trackers = issueTrackersForContext(context).filter((tracker) => tracker.type === type)
+  const requested = workspace.trim()
+  if (requested) {
+    return trackers.find((tracker) => tracker.workspace === requested)
+  }
+  return trackers[0]
```
AnaBerg: For GitHub Issues, picking the first tracker makes the Access step report the source as available when multiple workspace-managed trackers exist. The backend validates the same state as ambiguous and blocks save with `readiness-issue-tracker-ambiguous`, and this pattern currently has no UI field to select the tracker workspace. The wizard should surface the ambiguity or provide a selector instead of treating the first tracker as available.

cmd/workflow_setup.go
```diff
@@ -0,0 +1,350 @@
...
+func addWorkflowSetupFlags(cmd *cobra.Command, opts *workflowCreateOptions) {
+	cmd.Flags().StringVar(&opts.Workspace, "workspace", "", "workspace name [required]")
+	cmd.Flags().StringVar(&opts.Name, "name", "", "workflow name")
+	cmd.Flags().StringVar(&opts.Repository, "repo", "", "GitHub repository in owner/repo format for github-issue")
+	cmd.Flags().StringVar(&opts.Output, "output", "", "output YAML path")
+	cmd.Flags().BoolVar(&opts.Manual, "manual", true, "enable manual trigger inputs")
+	cmd.Flags().BoolVar(&opts.CreateWorkspace, "create-workspace", false, "create a minimal local workspace directory if missing")
+	cmd.Flags().StringVar(&opts.ConcurrencyGroup, "concurrency-group", "", "workflow concurrency group")
+	cmd.Flags().StringArrayVar(&opts.Labels, "label", nil, "GitHub issue label that can trigger the workflow (can be repeated)")
+	cmd.Flags().StringVar(&opts.IntegrationWorkspace, "integration-workspace", "", "issue tracker workspace for linear-status or shortcut-status")
```
AnaBerg, line 127: The CLI path cannot configure the advanced fields advertised by the pattern metadata and UI. For example, `workflow create github-issue ... --include-pre-commit` fails locally with `unknown flag`, so CLI-generated workflows cannot opt into the lifecycle feature this flow exposes. Consider adding explicit flags for pre-commit/status/done-signal fields or deriving supported CLI flags from the same pattern metadata.

pkg/workflowsetup/render.go
```diff
@@ -0,0 +1,609 @@
...
+
+func renderWorkflowName(requestedName string, fallback string) string {
+	name := strings.TrimSpace(requestedName)
+	if name == "" {
+		name = fallback
+	}
+	if err := ValidateSlug(name); err == nil {
+		return name
+	}
+	candidate := SlugCandidate(name)
```
AnaBerg, line 72: This silently rewrites any non-empty invalid workflow name into a slug. The UI currently only validates that the name is non-empty, so a user entering something like `Issue Precommit` gets `issue-precommit` without a field-level warning before save. That can be surprising and can hide name collisions. Consider validating the slug in the UI or returning an explicit warning when the backend rewrites the requested name.

cmd/factory.go
```diff
@@ -431,6 +434,227 @@ func runFactoryRm(name string) error {
...
+	})
+	if err != nil {
+		return nil, err
+	}
+	return files, nil
+}
+
+func printFactoryConvertResult(out io.Writer, result factoryConvertResult) {
+	if result.Status == workflowsetup.FactoryConvertStatusReady {
+		fmt.Fprintf(out, "Converted factory %q to workflow %q at %s\n", result.WorkflowName, result.WorkflowName, result.Path)
```
AnaBerg: This prints the generated workflow name in both placeholders, so the success message reports the wrong factory whenever the factory name and workflow name differ. Carry the original factory name in `factoryConvertResult`, or print the command input/factory name here.

pkg/hub/workflow_setup_api.go
```diff
@@ -0,0 +1,550 @@
...
+}
+
+func workflowSetupProviderReady(snapshot workflowsetup.SetupEnvironmentSnapshot) bool {
+	for _, provider := range snapshot.Providers {
+		if strings.TrimSpace(snapshot.DefaultProvider) != "" &&
+			strings.TrimSpace(provider.Name) != snapshot.DefaultProvider &&
+			strings.TrimSpace(provider.Type) != snapshot.DefaultProvider {
+			continue
+		}
+		if provider.Provisionable && provider.CredentialsSet {
```
AnaBerg: This still diverges from the actual readiness rules for `exedev`. `ValidateReadiness` now allows `exedev` without `SSHKeyPath`/`AccessToken` because the runtime can use the default SSH agent, and I reproduced that render/validate/save succeeds locally with `providers.exedev: {}`. However, the setup context reports `providerReady: false` here because `CredentialsSet` is false, so the wizard's Access step shows the runtime provider as missing even though the backend will save the workflow. This should use the same runtime-configured predicate as readiness, or otherwise special-case `exedev` consistently.

PR #388, opened 2026-06-09
Show dependency downtime banner

### LINKED ISSUE #260: show a banner when a dependent service is experiencing downtime
in the ui (and cli) we should be able to detect when one of the services (model, sandbox, issue tracker) is experiencing an issue. we are dealign with this a bit and it's hard to know that the problem just needs to wait

AnaBerg:
## Summary
- add an authenticated dependency status endpoint with cached external status checks
- surface downtime dependencies through the dashboard hook
- render a compact header banner with count, accessible label, and dependency list tooltip/fallback title

## Screenshots
image

## Validation
- go test ./pkg/hub -run TestDependencyStatus -count=1
- go test ./pkg/hub -count=1
- go test ./...
- cd web && npm run build
- cd web && npx eslint components/dependency-downtime-banner.tsx
- Browser check against a mocked hub response showing 2 dependencies with downtime

## Notes
- `npm run lint` for the whole web app still fails on existing React lint violations outside this change.
- The spec document remains uncommitted by request.

Closes #260


web/components/dependency-downtime-banner.tsx
```diff
@@ -0,0 +1,69 @@
...
+interface DependencyDowntimeBannerProps {
+  dependencies: DependencyStatus[]
+  className?: string
+}
+
+function bannerText(count: number): string {
+  return count === 1 ? "1 dependency with downtime" : `${count} dependencies with downtime`
+}
+
+function dependencyKindLabel(kind: string): string {
```
AnaBerg: would be best if `kind`was a enum

pkg/hub/dependency_status_test.go
```diff
@@ -0,0 +1,226 @@
...
+	}
+}
+
+func TestDependencyStatusFailureIsUnknownAndNotDowntime(t *testing.T) {
+	service := newDependencyStatusService(&types.HubConfig{
+		LLMKeys: types.LLMKeysList{{Name: "anthropic", Provider: "anthropic", APIKey: "sk-test"}},
+	})
+	service.checkers["model:anthropic"] = stubDependencyChecker{err: errors.New("status page unavailable")}
+
+	resp := service.snapshot(context.Background())
```
marccampbell, line 112: running this locally created a failure for me, i think its my configuration, but maybe we can harden this test case some.  i saw this test case report/check on github and linear because i had them set in my local hub config, which makes the test not-portable when running locally.

maybe we can change the implementation to accept the checks to run, instead of having it load them from config on disk?

marccampbell (approved):

PR #402, opened 2026-06-11
Add workflow artifact volumes

### LINKED ISSUE #392: Add workflow volumes backed by hub artifacts
## Summary

Add hub-managed volumes that workflows can attach to claws without committing the data to GitHub.

Volumes should use the artifact store from #390 for storage. They should support read-only shared access and single-writer read-write access with hub-managed leases. RWX/shared-writer support is explicitly out of scope for the first version.

Depends on #390. Checkpoint migration in #391 is a good first artifact-store consumer before this larger feature.

## Motivation

Some tasks need private or large filesystem state that should not be committed to the repository, such as:

- private test fixtures
- local databases or indexes
- generated caches
- model or tool data
- state shared across scheduled workflow runs

The agent should be able to use this data during a run while keeping it outside the Git working tree by default.

## Workflow shape

Example workflow configuration:

```yaml
volumes:
  - name: test-fixtures
    source: hub://volumes/test-fixtures
    mount: /mnt/elasticclaw/test-fixtures
    mode: ro

  - name: scratch-db
    source: hub://volumes/project-scratch
    mount: /mnt/elasticclaw/project-scratch
    mode: rw
```

Open question: volumes may be scoped to workspace, workflow, or run. Start with workspace-scoped persistent volumes unless implementation discovers a better local fit.

## Access model

Support two access modes initially:

- `ro`: many claws may attach concurrently, unless an active RW lease exists.
- `rw`: exactly one active claw may attach, and no RO leases may be active.

Do not support RWX/shared-writer semantics in the first version.

## Lease model

Keep locks/leases in the hub database. OCI/S3 artifact storage does not solve concurrency by itself.

Possible table:

```sql
volume_leases
- volume_id
- claw_id
- mode -- ro/rw
- manifest_digest
- acquired_at
- expires_at
- heartbeat_at
```

Rules:

- RO attach succeeds if there is no active RW lease.
- RW attach succeeds only if there are no active RO or RW leases.
- Leases are heartbeated by active claws.
- Hub releases leases on normal claw shutdown.
- Expired leases can be reclaimed.
- RW write-back should verify that the lease still owns the originally attached digest.

## Artifact flow

RO flow:

1. Resolve the volume ref to a manifest digest.
2. Pull/extract the artifact into the claw at the configured mount path.
3. Do not sync changes back on shutdown.

RW flow:

1. Acquire a RW lease.
2. Resolve and pin the current manifest digest.
3. Pull/extract the artifact into the claw at the configured mount path.
4. On shutdown, pack the volume directory and push a new artifact manifest.
5. Move the volume ref, such as `latest`, to the new manifest digest.
6. Release the lease.

## Storage format

Use archive transfer first for portability across providers:

- pack volume content as tar + zstd
- store as an ElasticClaw OCI-style artifact layer
- extract into a mount directory outside the Git repo

Provider-native volumes can be considered later as an optimization.

## Safety defaults

- Mount outside the repository by default, for example `/mnt/elasticclaw/<volume>`.
- Do not automatically commit volume contents.
- Require explicit workflow configuration for each attached volume.
- Make lease conflicts visible in workflow/run status.

## Acceptance criteria

- Workflows can declare volumes to attach to claws.
- RO volumes can be attached by multiple claws concurrently.
- RW volumes enforce a single active writer through hub leases.
- Volume contents are stored through the artifact store from #390.
- RW volumes sync back to a new artifact version on successful shutdown.
- Volume mounts default outside the Git repository.
- RWX/shared-writer support is not implemented and is documented as unsupported.

marccampbell:
## Summary
- add workflow `volumes:` schema for hub-managed artifact-backed directories
- add hub volume leases with shared RO and single-writer RW semantics
- add claw-bridge attach/sync messages and artifact archive endpoints for volume transfer
- sync RW volumes before hub-driven terminal shutdown and heartbeat active leases

## Notes
- Uses tar+gzip archives for the first portable volume layer implementation. The artifact layer media type is explicit so zstd can be added later without changing the lease model.
- RWX/shared-writer semantics are intentionally not implemented.

Closes #392

## Tests

### DIFF
--- a/pkg/hub/volumes.go
+++ b/pkg/hub/volumes.go
@@ -208,6 +208,12 @@ func (s *Server) heartbeatWorkflowVolumeLeases(clawID string) {
 	_, _ = s.db.Exec(`UPDATE volume_leases SET heartbeat_at=?, expires_at=? WHERE claw_id=? AND released_at IS NULL`, now, now.Add(volumeLeaseTTL), clawID)
 }
 
+func (s *Server) workflowVolumeLeaseActive(leaseID string) bool {
+	var active int
+	err := s.db.QueryRow(`SELECT COUNT(*) FROM volume_leases WHERE id=? AND released_at IS NULL AND expires_at > ?`, leaseID, time.Now().UTC()).Scan(&active)
+	return err == nil && active > 0
+}
+
 func (s *Server) workflowVolumesForClaw(clawID string) ([]workflowVolumeRuntime, error) {
 	var raw string
 	if err := s.db.QueryRow(`SELECT COALESCE(workflow_volumes,'[]') FROM claws WHERE id=?`, clawID).Scan(&raw); err != nil {
@@ -296,10 +302,10 @@ func (s *Server) syncWorkflowVolumes(clawID string) {
 	if err != nil || len(volumes) == 0 {
 		return
 	}
-	hasWritableLease := false
+	hasActiveWritableLease := false
 	for _, volume := range volumes {
-		if volume.Mode == volumeModeRW && volume.LeaseID != "" {
-			hasWritableLease = true
+		if volume.Mode == volumeModeRW && volume.LeaseID != "" && s.workflowVolumeLeaseActive(volume.LeaseID) {
+			hasActiveWritableLease = true
 			break
 		}
 	}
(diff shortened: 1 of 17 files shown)


pkg/hub/volumes.go
```diff
@@ -0,0 +1,473 @@
...
+	s.mu.RUnlock()
+	if cc == nil {
+		s.releaseWorkflowVolumeLeases(clawID)
+		return
+	}
+	for _, volume := range volumes {
+		if volume.Mode != volumeModeRW || volume.LeaseID == "" {
+			continue
+		}
+		if err := s.dispatchVolumeSync(context.Background(), cc, volume); err != nil {
```
AnaBerg, line 339: Should the lease still be released if `dispatchVolumeSync` fails? If the upload fails here, releasing the lease may allow another run to continue from the previous volume state while this claw's changes were not saved.

PR #413, opened 2026-06-20
Add Jira issue tracker support

marccampbell:
## Summary
- add Jira as a first-class issue tracker with workflow schema, managed tracker settings, DB storage, webhooks, polling recovery, status transitions, comments, and token injection
- add Jira tracker configuration to the settings UI and webhook copy list
- document the intended gated Jira Software container E2E path

## Tests
- env GOCACHE=/private/tmp/elasticclaw-go-build go test ./pkg/types ./pkg/hub
- npm run build

## Notes
- Jira webhook validation supports a shared secret via the `secret` query parameter or `X-ElasticClaw-Webhook-Secret` header because Jira Data Center does not expose the same HMAC webhook contract as GitHub/Shortcut.
- Container-backed Jira E2E remains a follow-up gated job due Jira bootstrap/licensing/startup cost.

### DIFF
--- a/pkg/hub/linear.go
+++ b/pkg/hub/linear.go
@@ -1295,9 +1295,6 @@ func (s *Server) handleClawDoneSignal(clawID, rawMessage string) {
 			if ghToken != "" {
 				// Parse gh-owner/repo/number from issueID
 				rest := strings.TrimPrefix(issueID, "gh-")
-				if rest == issueID {
-					rest = issueID
-				}
 				lastSlash := strings.LastIndex(rest, "/")
 				if lastSlash > 0 {
 					repo := rest[:lastSlash]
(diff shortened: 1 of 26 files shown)


pkg/hub/linear.go
```diff
@@ -1378,7 +1421,7 @@ func (s *Server) completeNoPRDoneClaw(clawID, tenantID, issueID string) {
 func (s *Server) handleClawTerminateSignal(clawID, rawMessage string) {
```
AnaBerg, line 1402: this function posts self-termination comments for tracker types like GitHub Issues but is missing Jira handling

marccampbell: thanks, resolved in a new commit in this PR

PR #436, opened 2026-07-07
Stop swallowing migration errors

AnaBerg:
## Summary

Implements docs/re-arch/phase-0-stop-the-bleeding.md item 0.5.

`pkg/hub/db.go` ran ~30 `ALTER TABLE ... ADD COLUMN` statements with `_, _ = db.Exec(...)`, so a real migration failure (full disk, lock, permission denied) was invisible and boot would continue against a broken schema.

## Changes

- Added `execIgnoreDuplicate(db *sql.DB, stmt string) error`: runs the statement and classifies the result.
  - `"duplicate column name"` (column already exists on a re-run) → ignored.
  - `"no such table"` (table not created yet on a brand-new DB — it's created later in the same `migrate()` call) → ignored.
  - Any other error is logged (`[db] migration failed: ...`) and returned, wrapped as `migrate: %w`, which aborts `openDB`/boot.
- Replaced every previously-swallowed `ALTER TABLE ... ADD COLUMN` call in `migrate()` (claws, claw_prs, messages, factory_triggers, volume_leases) with calls through `execIgnoreDuplicate`, looping over slices of statements to keep the diff readable.
- Left the `CREATE TABLE IF NOT EXISTS` / `CREATE INDEX IF NOT EXISTS` / data-backfill statements untouched — those already use idempotent SQL (`IF NOT EXISTS`, `INSERT OR IGNORE`) and aren't part of this item's scope.
- No change to the migration mechanism itself (no versioning) — that's phase 1.4.

## Verification

- `go build ./...` — passes.
- `go test ./pkg/hub/ -count=1` — passes (no `-tags integration`).
- Added two tests in `pkg/hub/db_test.go`:
  - `TestMigrateIdempotent`: runs `migrate()` twice against the same DB and asserts no error — confirms normal boot stays idempotent after re-running against an already-migrated schema.
  - `TestMigrateAbortsOnRealError`: creates a bare `claws` table, chmods the DB file read-only, opens it in `mode=ro`, and asserts `migrate()` returns a wrapped `migrate: ...` error (observed: `attempt to write a readonly database (8)`) instead of silently succeeding — this stands in for the full-disk/lock scenario from the spec.

```
$ go build ./...
(no output — success)

$ go test ./pkg/hub/ -count=1 -v -run 'TestMigrate'
=== RUN   TestMigrateIdempotent
--- PASS: TestMigrateIdempotent (0.03s)
=== RUN   TestMigrateAbortsOnRealError
2026/07/07 16:57:38 [db] migration failed: attempt to write a readonly database (8)
--- PASS: TestMigrateAbortsOnRealError (0.00s)
PASS

$ go test ./pkg/hub/ -count=1
ok  	github.com/elasticclaw/elasticclaw/pkg/hub	14.873s
```

### DIFF
--- a/pkg/hub/db.go
+++ b/pkg/hub/db.go
@@ -4,10 +4,36 @@ import (
 	"database/sql"
 	"fmt"
 	"log"
+	"strings"
 	"time"
 
 	_ "modernc.org/sqlite" // pure-Go SQLite, no CGO required
 )
+
+// execIgnoreDuplicate runs stmt (expected to be an ALTER TABLE ... ADD COLUMN)
+// and ignores errors that are expected outcomes of this migration style rather
+// than real failures:
+//   - "duplicate column name": the column already exists — SQLite has no
+//     ALTER TABLE ... ADD COLUMN IF NOT EXISTS, so re-running against an
+//     already-migrated database always hits this.
+//   - "no such table": the table itself doesn't exist yet on a brand-new
+//     database — it gets created later in this same migration by the
+//     CREATE TABLE IF NOT EXISTS block below.
+//
+// Any other error (full disk, lock, permission denied, ...) is returned so the
+// caller can abort boot instead of silently continuing with a broken schema.
+func execIgnoreDuplicate(db *sql.DB, stmt string) error {
+	_, err := db.Exec(stmt)
+	if err == nil {
+		return nil
+	}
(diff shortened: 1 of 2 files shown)


AnaBerg:
Closing: this implementation round was cancelled; only the docs/re-arch specs branch is being kept.

PR #439, opened 2026-07-07
feat(hub): restrict CORS to configured allowed origins

AnaBerg:
Implements item **0.4 Restricted CORS** of the Phase 0 ("stop the bleeding") re-architecture plan.

## Motivation

`corsMiddleware` in `pkg/hub/server.go` unconditionally set `Access-Control-Allow-Origin: *` (plus permissive `Allow-Headers`/`Allow-Methods`) on every response, for any origin, and never sent `Vary: Origin`. Any website could issue cross-origin requests against the hub API, and shared caches could serve a CORS-decorated response to the wrong origin. Per the spec, the allowed origin must come from hub.yaml config (`allowed_origins`), defaulting to the hub's own origin, and `Vary: Origin` must always be sent.

## Dependencies

## What changed

- `pkg/types/template.go`: new `AllowedOrigins []string` field on `HubConfig` (yaml: `allowed_origins`). The config loader uses `KnownFields(true)`, so the struct field is required for the key to be accepted.
- `pkg/hub/server.go`:
  - `corsMiddleware` now takes an allowlist. It echoes the request `Origin` back (with the `Allow-Headers`/`Allow-Methods` headers) only when the normalized origin is on the list; otherwise it omits all `Access-Control-*` headers. `Vary: Origin` is always added. Preflight `OPTIONS` still short-circuits with 204.
  - New `allowedCORSOrigins()` builds the set from `allowed_origins`; when the list is empty it defaults to the hub's own origin derived from `url` and `public_url`. Invalid entries are logged and skipped. Matching is case-insensitive with trailing slashes stripped.
- `pkg/hub/cors_test.go`: unit tests for allowed/disallowed/no-origin requests, preflight, case-insensitive matching, config parsing, and the own-origin default.
- `docker/hub.dev.yaml.example`: documents `allowed_origins` with `http://localhost:3000` so the dev compose (web:3000 → hub:8080) keeps working.

## Testing

- `go build ./...` — pass.
- `make test` equivalent (`go test ./...`) — all packages pass, including `pkg/hub` (16.6s) and `pkg/hub/factorytest`.
- `go test ./pkg/hub/ -run 'TestCORS|TestAllowedCORS' -v` — 7/7 pass.

Note for deployers: this is a behavior change — cross-origin browser clients not covered by the default (hub's own `url`/`public_url` origin) now need an explicit `allowed_origins` entry in hub.yaml. Dev setups must copy the updated `docker/hub.dev.yaml.example` entry.

### DIFF
--- a/pkg/hub/cors_test.go
+++ b/pkg/hub/cors_test.go
@@ -125,15 +125,31 @@ func TestAllowedCORSOriginsFromConfig(t *testing.T) {
 func TestAllowedCORSOriginsDefaultsToHubOrigin(t *testing.T) {
 	s := &Server{hubCfg: &types.HubConfig{
 		URL:       "http://localhost:8080/some/path",
-		PublicURL: "https://hub.example.com",
+		PublicURL: "http://hub:8080",
 	}}
 	got := s.allowedCORSOrigins()
-	for _, want := range []string{"http://localhost:8080", "https://hub.example.com"} {
-		if _, ok := got[want]; !ok {
-			t.Errorf("allowedCORSOrigins missing default %q (got %v)", want, got)
-		}
+	if _, ok := got["http://localhost:8080"]; !ok {
+		t.Errorf("allowedCORSOrigins missing default hub origin (got %v)", got)
 	}
-	if len(got) != 2 {
-		t.Errorf("allowedCORSOrigins has %d entries, want 2: %v", len(got), got)
+	if _, ok := got["http://hub:8080"]; ok {
+		t.Error("public_url (claw callback URL) must not be implicitly allowed as a browser origin")
+	}
+	if len(got) != 1 {
+		t.Errorf("allowedCORSOrigins has %d entries, want 1: %v", len(got), got)
+	}
+}
+
+func TestAllowedCORSOriginsPublicURLRequiresExplicitListing(t *testing.T) {
+	s := &Server{hubCfg: &types.HubConfig{
+		URL:            "http://localhost:8080",
+		PublicURL:      "https://hub.example.com",
+		AllowedOrigins: []string{"https://hub.example.com"},
+	}}
+	got := s.allowedCORSOrigins()
+	if _, ok := got["https://hub.example.com"]; !ok {
+		t.Errorf("explicitly listed public_url origin must be allowed (got %v)", got)
+	}
+	if len(got) != 1 {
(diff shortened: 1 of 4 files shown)


pkg/hub/server.go
```diff
@@ -409,6 +416,39 @@ func corsMiddleware(next http.Handler) http.Handler {
...
+	}
+	if len(s.hubCfg.AllowedOrigins) > 0 {
+		for _, o := range s.hubCfg.AllowedOrigins {
+			add(o)
+		}
+		return allowed
+	}
+	// Default: only the hub's own origin.
+	add(s.hubCfg.URL)
+	add(s.hubCfg.PublicURL)
```
AnaBerg: docs/re-arch spec-alignment: Phase 0.4 says the omitted-config default should be the hub's own browser origin. `public_url` is documented and used as the claw/agent callback URL, and the dev example sets it to the Docker-internal `http://hub:8080`; allowing it by default broadens browser CORS beyond the actual UI/API origin. Please default only from the browser-facing hub URL, or require `public_url` to be explicitly listed in `allowed_origins` when it is intentionally a browser origin.

pkg/hub/cors_test.go
```diff
@@ -0,0 +1,155 @@
...
+	}
+	if got := rec.Header().Get("Vary"); got != "Origin" {
+		t.Errorf("Vary = %q, want %q", got, "Origin")
+	}
+}
+
+func TestCORSMiddlewarePreflight(t *testing.T) {
+	h := newCORSHandler(map[string]struct{}{"http://localhost:3000": {}})
+
+	rec := doCORSRequest(t, h, http.MethodOptions, "http://localhost:3000")
```
AnaBerg: This does not model an actual browser preflight because the request omits `Access-Control-Request-Method` (and requested headers where relevant). As written, the test also permits the middleware to short-circuit every `OPTIONS` request as 204, including non-CORS `OPTIONS`, which is what the current implementation does. Please set the preflight headers in these cases and add coverage showing that an ordinary `OPTIONS` request is delegated to the next handler.

AnaBerg:
The common execution rules in `docs/re-arch` require a CHANGELOG/release note for observable CORS behavior changes. This PR changes which browser origins can call the hub, but its diff contains no changelog or release-note entry. Please document the new `allowed_origins` requirement and the upgrade behavior for existing cross-origin deployments.

PR #441, opened 2026-07-07
feat(hub): recovery middleware and unified JSON error envelope

AnaBerg:
Implements item **0.2 Recovery middleware + error envelope** of the Phase 0 ("stop the bleeding") re-architecture plan.

## Motivation

No handler recovers from panics, and error responses alternate between plain-text http.Error and ad-hoc JSON; the UI needs a single machine-readable error envelope.

Concrete evidence in the code:

- `pkg/hub/server.go` booted with `http.ListenAndServe(s.addr, corsMiddleware(mux))` — nothing in the chain recovers, so a panic in any of the 100+ registered handlers kills the request goroutine, and handlers holding `s.mu` or mid-write can leave shared state corrupted.
- Error responses were inconsistent even within a single handler: `handleClawDetail` used plain-text `http.Error(w, "db error", ...)` while other paths used the ad-hoc `jsonError` helper that emitted `{"error": msg}` with no stable code.
- The frontend (`web/lib/api.ts:78-106`) already tries `JSON.parse` on error bodies and reads `parsed.error`, falling back to raw text — so a gradual migration to a JSON envelope is backward-compatible.

## Dependencies

Note: the recovery log includes method/path/remote addr only; request IDs arrive in Phase 1, and the remaining (non-UI) handlers migrate to the envelope in Phase 2 together with the `pkg/hub` reorganization, per the spec.

## What changed

- New `pkg/hub/middleware.go`:
  - `apiError` envelope (`{"error": "...", "code": "..."}`, `code` omitted when empty) and `writeErr(w, status, code, msg)` helper.
  - `withRecovery` middleware: recovers panics, logs the stack with method/path/remote addr, responds 500 with the JSON envelope. Re-panics on `http.ErrAbortHandler` to preserve the stdlib contract.
- Wired `withRecovery` at the top of the chain (before CORS) in `Server.Run`.
- Migrated only the handlers the UI uses, per the spec — `/api/claws*` (incl. subresources `prs`/`checkpoints`), `/api/messages*`, `/api/settings*` (incl. `ai-config` and `model-auth`), `/api/auth/*` — replacing `http.Error` with `writeErr` (122 call sites across `server.go`, `settings.go`, `ai_config.go`, `model_auth.go`, `auth_github.go`, `pr_watcher.go`, `checkpoints.go`). Status-derived stable codes: `bad_request`, `unauthorized`, `forbidden`, `not_found`, `method_not_allowed`, `conflict`, `internal`, `upstream_error`, `unprocessable`.
- `jsonError` now delegates to `writeErr` so both paths emit the identical envelope.
- New `pkg/hub/middleware_test.go`: injects a panic into a fake handler and verifies 500 + JSON envelope + logged stack (spec acceptance criterion), plus passthrough, `ErrAbortHandler` re-panic, and `writeErr` encoding tests.

## Testing

- `go build ./...` — pass.
- `go test ./...` (same as `make test` minus `-v`) — all packages pass, including `pkg/hub` (16s) and `pkg/hub/factorytest`.
- `go test ./pkg/hub/ -run "TestWithRecovery|TestWriteErr" -v` — 5/5 pass.

### DIFF
--- a/pkg/hub/model_auth.go
+++ b/pkg/hub/model_auth.go
@@ -56,7 +56,7 @@ const (
 
 func (s *Server) handleModelAuthLogin(w http.ResponseWriter, r *http.Request) {
 	if r.Method != http.MethodPost {
-		http.Error(w, "method not allowed", http.StatusMethodNotAllowed)
+		writeErr(w, http.StatusMethodNotAllowed, "method_not_allowed", "method not allowed")
 		return
 	}
 	var req struct {
@@ -65,7 +65,7 @@ func (s *Server) handleModelAuthLogin(w http.ResponseWriter, r *http.Request) {
 		Mode     string `json:"mode"`
 	}
 	if err := json.NewDecoder(r.Body).Decode(&req); err != nil {
-		http.Error(w, "invalid request", http.StatusBadRequest)
+		writeErr(w, http.StatusBadRequest, "bad_request", "invalid request")
 		return
 	}
 	req.Provider = strings.TrimSpace(req.Provider)
@@ -77,11 +77,11 @@ func (s *Server) handleModelAuthLogin(w http.ResponseWriter, r *http.Request) {
 		req.Profile = req.Provider + "-default"
 	}
 	if req.Provider != "codex" && req.Provider != "grok" {
-		http.Error(w, "provider must be codex or grok", http.StatusBadRequest)
+		writeErr(w, http.StatusBadRequest, "bad_request", "provider must be codex or grok")
 		return
 	}
 	if req.Mode != "device" {
-		http.Error(w, "only device login is supported", http.StatusBadRequest)
+		writeErr(w, http.StatusBadRequest, "bad_request", "only device login is supported")
 		return
 	}
 
@@ -107,7 +107,7 @@ func (s *Server) handleModelAuthLogin(w http.ResponseWriter, r *http.Request) {
 
 func (s *Server) handleModelAuthLoginStatus(w http.ResponseWriter, r *http.Request) {
 	if r.Method != http.MethodGet {
-		http.Error(w, "method not allowed", http.StatusMethodNotAllowed)
+		writeErr(w, http.StatusMethodNotAllowed, "method_not_allowed", "method not allowed")
 		return
 	}
 	id := r.PathValue("id")
--- a/pkg/hub/model_auth.go
+++ b/pkg/hub/model_auth.go
@@ -116,7 +116,7 @@ func (s *Server) handleModelAuthLoginStatus(w http.ResponseWriter, r *http.Reque
 	snapshot := snapshotModelAuthLoginJob(job)
 	s.modelAuthJobsMu.Unlock()
 	if job == nil {
-		http.NotFound(w, r)
+		writeErr(w, http.StatusNotFound, "not_found", "not found")
 		return
 	}
 	jsonOK(w, snapshot)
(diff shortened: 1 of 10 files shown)


pkg/hub/server.go
```diff
@@ -268,7 +268,7 @@ func (s *Server) Run(opts ...RunOptions) error {
 	if s.hubCfg.UIPassword == "" {
```
AnaBerg, line 268: [docs/re-arch spec-alignment] Phase 0.2 says UI handlers under `/api/claws*`, `/api/messages*`, `/api/settings*`, and `/api/auth/*` should return the JSON error envelope. The route chain now has recovery, and many handlers use `writeErr`, but the auth middleware that wraps these same UI routes still returns `http.Error` for unauthorized/forbidden cases (`withAuth`, `withWebAuth`, `withWebAdminAuth`, and `withConfigMutationAuth`). As a result, common UI errors like an expired session on `/api/auth/me`, `/api/settings`, `/api/claws`, or `/api/messages` remain `text/plain` instead of `{ "error": ..., "code": ... }`. Please migrate those middleware error paths to `writeErr` as part of this phase.

pkg/hub/model_auth.go
```diff
@@ -107,7 +107,7 @@ func (s *Server) handleModelAuthLogin(w http.ResponseWriter, r *http.Request) {
 
 func (s *Server) handleModelAuthLoginStatus(w http.ResponseWriter, r *http.Request) {
 	if r.Method != http.MethodGet {
-		http.Error(w, "method not allowed", http.StatusMethodNotAllowed)
+		writeErr(w, http.StatusMethodNotAllowed, "method_not_allowed", "method not allowed")
```
AnaBerg, line 110: [docs/re-arch spec-alignment] This settings UI endpoint was partially migrated to `writeErr`, but a missing model-auth job still falls through to `http.NotFound` below. The frontend polls `/api/settings/model-auth/login/{id}`, so that error path is in the phase 0.2 UI-handler scope and should use the same envelope, e.g. `writeErr(w, http.StatusNotFound, "not_found", "not found")`, with a test for the missing-job case.

pkg/hub/server.go
```diff
@@ -1688,7 +1688,7 @@ func hiddenSystemMessagesSQL() string {
 
 func (s *Server) handleMessageTimeline(w http.ResponseWriter, r *http.Request, tenantID, clawID string) {
 	if r.Method != http.MethodGet {
-		http.Error(w, "method not allowed", http.StatusMethodNotAllowed)
+		writeErr(w, http.StatusMethodNotAllowed, "method_not_allowed", "method not allowed")
 		return
 	}
 	if !s.canViewMessages(w, r, tenantID, clawID) {
```
AnaBerg, line 1694: manual review: `handleMessageTimeline` and `handleMessageActivity` now use `writeErr`, but both still return through `canViewMessages` before those migrated error paths. That helper still calls `http.Error` for not-found/db/forbidden cases, so GitHub OAuth sessions hitting `/api/messages/{id}/timeline` or `/api/messages/{id}/activity` without access (or for a missing claw) still get `text/plain` instead of the phase-0 JSON envelope. Please migrate `canViewMessages` to `writeErr` too, and add a test for one of these OAuth-denied or missing-claw paths.

AnaBerg:
The recovery middleware writes directly to the original `ResponseWriter`, so a panic after a handler has already called `WriteHeader`/`Write` cannot produce the required 500 response: the client keeps the original status (often 200) and receives a JSON error appended to a partial body. The implementation even acknowledges this as a best-effort superfluous `WriteHeader`. This does not satisfy the Phase 0.2 contract that recovered panics respond with the 500 error envelope. Please either buffer API responses until the handler completes or explicitly handle the committed-response case (for example by aborting the connection instead of appending a misleading envelope).

PR #442, opened 2026-07-07
feat(hub): graceful shutdown with root context and http.Server

AnaBerg:
Implements item **0.1 Graceful shutdown** of the Phase 0 ("stop the bleeding") re-architecture plan.

## Motivation

The hub boots with bare `http.ListenAndServe` and 8+ background goroutines with no cancellation; SIGTERM kills the process mid-SQLite-write and drops WS connections without a goodbye.

- `pkg/hub/server.go` `Run()` ended in `return http.ListenAndServe(s.addr, corsMiddleware(mux))` — no timeouts, no shutdown path at all.
- `NewServer` fired off `go srv.pollProviderStatus()`, `go srv.keepAliveDaytonaSandboxes()`, `go srv.pruneAnalytics()`, `go srv.statusWatchdog()`, `go srv.checkpointScheduler()`, plus the PR watcher (`startPRWatcher`), the cron scheduler, and the integration poller (`startIntegrationPoller`). Every one was a `for range ticker.C` loop with no context and no way to stop — the cron scheduler even had a `stop()` method that nothing ever called.
- `pollAllPRs` in `pr_watcher.go` already contains a workaround for the missing shutdown ordering: it string-matches `"database is closed"` on query errors, i.e. pollers were known to race a dying DB.
- `/healthz` unconditionally returned 200, so a load balancer would keep routing traffic to a process that was already being killed.

## Dependencies

Note: the spec's execution rules ask for a CHANGELOG note for observable-behavior changes, but the repo has no CHANGELOG file yet; the behavior change is documented here and in the `shutdown_grace` config comment instead.

## What changed

- **Root context + signal handling** (`pkg/hub/server.go`): `Run()` creates a root context via `signal.NotifyContext(context.Background(), os.Interrupt, syscall.SIGTERM)` and blocks until either the listener fails or a signal arrives.
- **Explicit `http.Server`**: replaces `http.ListenAndServe` with a server configured with `ReadHeaderTimeout: 10s` and `IdleTimeout: 120s`; on signal, `srv.Shutdown(drainCtx)` drains in-flight requests for up to 15s, configurable via a new `shutdown_grace` field in hub.yaml (`types.HubConfig`).
- **Cancellable background goroutines**: all seven ticker loops (`pollProviderStatus`, `keepAliveDaytonaSandboxes`, `pruneAnalytics`, `statusWatchdog`, `checkpointScheduler`, PR watcher — now `watchPRs` —, integration poller — now `pollIntegrationsLoop`) take a `context.Context` and exit on `ctx.Done()`. They run under a `golang.org/x/sync/errgroup.Group` (already a dependency) so shutdown can wait for all of them.
- **`Server.Shutdown(ctx)`** implements the specced order: mark draining → stop accepting / drain HTTP → close claw WebSocket connections (main + status channel) with a `StatusGoingAway` close frame — `"killed"` is NOT sent, so sandboxes stay alive and bridges reconnect when the hub returns → cancel root context and wait for the errgroup → stop the cron scheduler (its existing `stop()` is finally called) → close `sql.DB`. Guarded by `sync.Once` so repeated calls are safe.
- **`/healthz` returns 503 while draining** (was: always 200).
- **Test helpers** (`pkg/hub/testhelpers.go`): the test server gets the same context/errgroup wiring, cancelled via `t.Cleanup`, so `StartPRWatcherForTest` goroutines no longer leak across tests.

## Testing

- `go build ./...` — pass.
- `go vet ./pkg/hub/...` — pass.
- `go test ./...` — all packages pass (`pkg/hub` in 16.1s, plus cmd, claw-bridge, providers, e2e).
- `make test-factory` — pass.
- New tests in `pkg/hub/shutdown_test.go`:
  - `TestGracefulShutdown` boots a full `NewServer` (all background goroutines + cron) and verifies `Shutdown` returns within the drain window, cancels the root context, leaves no background goroutine running (`bg.Wait` returns immediately), closes the DB, and that a second `Shutdown` call is a safe no-op.
  - `TestHealthzReturns503WhileDraining` verifies `/healthz` flips 200 → 503 once draining starts.
  - `TestShutdownGraceConfig` covers default/valid/invalid/negative `shutdown_grace` values.

### DIFF
--- a/pkg/hub/checkpoints.go
+++ b/pkg/hub/checkpoints.go
@@ -122,11 +122,16 @@ func checkpointManifestPath(id string) string {
 	return filepath.Join(checkpointsRoot(), "manifests", id+".json")
 }
 
-func (s *Server) checkpointScheduler() {
+func (s *Server) checkpointScheduler(ctx context.Context) {
 	ticker := time.NewTicker(time.Minute)
 	defer ticker.Stop()
-	for range ticker.C {
-		s.requestIdleCheckpoints()
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case <-ticker.C:
+			s.requestIdleCheckpoints()
+		}
 	}
 }
 
--- a/pkg/hub/checkpoints.go
+++ b/pkg/hub/checkpoints.go
@@ -130,12 +130,16 @@ func (s *Server) checkpointScheduler(ctx context.Context) {
 		case <-ctx.Done():
 			return
 		case <-ticker.C:
-			s.requestIdleCheckpoints()
+			s.requestIdleCheckpoints(ctx)
 		}
 	}
 }
 
-func (s *Server) requestIdleCheckpoints() {
+// requestIdleCheckpoints asks idle claws for a checkpoint. Each request runs in
+// its own goroutine tracked by s.bg and receives the scheduler's context, so
+// shutdown both cancels in-flight requests and waits for them before closing
+// the database.
+func (s *Server) requestIdleCheckpoints(ctx context.Context) {
 	now := time.Now()
 	s.mu.RLock()
 	items := make([]struct {
@@ -162,11 +166,13 @@ func (s *Server) requestIdleCheckpoints() {
 		if s.hasRecentCheckpoint(item.id, checkpointMinInterval) {
 			continue
 		}
-		go func(clawID string) {
-			if _, err := s.requestCheckpoint(context.Background(), clawID, "idle-timer", "hub", false, checkpointRequestTimeout); err != nil {
+		clawID := item.id
+		s.bg.Go(func() error {
+			if _, err := s.requestCheckpoint(ctx, clawID, "idle-timer", "hub", false, checkpointRequestTimeout); err != nil {
 				log.Printf("[checkpoint] idle request for %s failed: %v", shortID(clawID), err)
 			}
-		}(item.id)
+			return nil
+		})
 	}
 }
 
--- a/pkg/hub/server.go
+++ b/pkg/hub/server.go
@@ -7,6 +7,7 @@ import (
 	"database/sql"
 	"encoding/base64"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"io/fs"
@@ -16,12 +17,15 @@ import (
 	"net/url"
 	"os"
 	"os/exec"
+	"os/signal"
 	"path"
 	"path/filepath"
 	"sort"
 	"strconv"
 	"strings"
 	"sync"
+	"sync/atomic"
+	"syscall"
 	"time"
 
 	"github.com/elasticclaw/elasticclaw/internal/webui"
@@ -35,6 +39,7 @@ import (
 	"github.com/elasticclaw/elasticclaw/pkg/types"
 	"github.com/google/uuid"
 	gossh "golang.org/x/crypto/ssh"
+	"golang.org/x/sync/errgroup"
 	"nhooyr.io/websocket"
 	"nhooyr.io/websocket/wsjson"
 )
@@ -92,6 +97,19 @@ type Server struct {
 
 	// cronScheduler manages scheduled workflow runs
 	cronScheduler *cronScheduler
+
+	// Graceful shutdown machinery.
+	// bgCtx is the root context handed to every background goroutine; bgCancel
+	// cancels it during shutdown. bg waits for all background goroutines.
+	bgCtx    context.Context
+	bgCancel context.CancelFunc
+	bg       *errgroup.Group
+	// httpSrv is the HTTP server started by Run (nil in tests that use Handler()).
+	httpSrv *http.Server
+	// draining is set once shutdown starts; /healthz returns 503 while draining.
+	draining atomic.Bool
+	// shutdownOnce guards Shutdown so it runs at most once.
+	shutdownOnce sync.Once
 }
 
 type clawConn struct {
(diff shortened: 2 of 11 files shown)


pkg/hub/server.go
```diff
@@ -221,20 +239,24 @@ func NewServer(addr, dbPath, identityDir string, hubCfg *types.HubConfig) (*Serv
...
+
+	// Start background pollers; each one exits when bgCtx is cancelled and is
+	// waited on by Shutdown via the errgroup.
+	srv.bg.Go(func() error { srv.pollProviderStatus(srv.bgCtx); return nil })
+	srv.bg.Go(func() error { srv.keepAliveDaytonaSandboxes(srv.bgCtx); return nil })
+	srv.bg.Go(func() error { srv.pruneAnalytics(srv.bgCtx); return nil })
+	srv.bg.Go(func() error { srv.statusWatchdog(srv.bgCtx); return nil })
+	srv.bg.Go(func() error { srv.checkpointScheduler(srv.bgCtx); return nil })
+	srv.bg.Go(func() error { srv.watchPRs(srv.bgCtx); return nil })
+	srv.bg.Go(func() error { srv.pollIntegrationsLoop(srv.bgCtx); return nil })
```
AnaBerg, line 253: docs/re-arch spec-alignment: Phase 0.1 explicitly lists `cron` among the background goroutines that should receive the root context and be waited via the shutdown group, but this new `bg.Go` set still omits `cronScheduler`. The scheduler is started outside `bgCtx` below and stopped later with `cron.Stop()`, which waits for running jobs that still use background contexts, so an active scheduled workflow can keep running past the drain window and continue touching the DB during shutdown. Please wire the cron scheduler/jobs to the same root context and include it in the errgroup shutdown path, or otherwise make `stop` bounded by the drain context.

pkg/hub/checkpoints.go
```diff
@@ -122,11 +122,16 @@ func checkpointManifestPath(id string) string {
...
 	ticker := time.NewTicker(time.Minute)
 	defer ticker.Stop()
-	for range ticker.C {
-		s.requestIdleCheckpoints()
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case <-ticker.C:
+			s.requestIdleCheckpoints()
```
AnaBerg: docs/re-arch spec-alignment: this makes the scheduler loop cancellable, but `requestIdleCheckpoints()` still starts per-claw checkpoint goroutines with `context.Background()` and they are not tracked by `s.bg`. That means `checkpointScheduler` can return and `s.bg.Wait()` can finish while an idle checkpoint is still writing SQLite, after which `Shutdown` may close the DB underneath it. Please pass the scheduler/root context into the checkpoint request path and run these spawned requests under a tracked errgroup, or avoid spawning untracked work during shutdown.

pkg/hub/server.go
```diff
@@ -268,7 +290,113 @@ func (s *Server) Run(opts ...RunOptions) error {
...
+				err = shutdownErr
+			}
+		}
+
+		log.Printf("[hub] closing claw connections")
+		s.closeClawConnections()
+
+		log.Printf("[hub] stopping background goroutines")
+		s.bgCancel()
+		_ = s.bg.Wait()
```
AnaBerg: docs/re-arch spec-alignment: `Shutdown(ctx)` receives the drain deadline, but this wait ignores it. Several tick bodies reached by the new context-aware loops still perform uncancelled `http.DefaultClient.Do` calls or other work that does not receive `bgCtx`, so a stuck poll can block here indefinitely and violate the Phase 0.1 acceptance criterion that SIGTERM exits within the drain window. Please either propagate `bgCtx` into the tick bodies/API requests so they unblock on shutdown, or wait for `s.bg.Wait()` in a goroutine and return `ctx.Err()` when the drain context expires.

pkg/hub/server.go
```diff
@@ -268,7 +290,113 @@ func (s *Server) Run(opts ...RunOptions) error {
...
+//
+// It is safe to call Shutdown multiple times; only the first call runs.
+func (s *Server) Shutdown(ctx context.Context) error {
+	var err error
+	s.shutdownOnce.Do(func() {
+		s.draining.Store(true)
+
+		if s.httpSrv != nil {
+			log.Printf("[hub] draining HTTP connections")
+			if shutdownErr := s.httpSrv.Shutdown(ctx); shutdownErr != nil && !errors.Is(shutdownErr, http.ErrServerClosed) {
```
AnaBerg: manual review: `httpSrv.Shutdown(ctx)` blocks until active HTTP handlers finish or the drain context expires, so this can spend the entire `shutdown_grace` before `closeClawConnections()` sends the claw close frames. That reverses the Phase 0.1 order (`stop accepting connections -> close claw WS connections -> wait goroutines -> close DB`) and means a long REST request can keep bridges connected until the drain deadline, then the code continues toward DB close with no budget left. Please split "stop accepting" from the blocking HTTP drain, or start the HTTP drain and close the claw/status websockets before waiting for in-flight handlers.

AnaBerg:
Manual validation against `docs/re-arch` Phase 0 item 0.1 found a blocker.

Core shutdown behavior is partially working: focused shutdown tests, `make test`, factory tests, and a manual SIGTERM run with a connected claw all passed in the sense that the process exited quickly, the WS received a `1001` close frame, logs showed the drain sequence, and no `hub.db-wal`/`hub.db-shm` remained.

However, the PR does not yet fully satisfy the spec requirement that background work is tied to the root context and waited before `sql.DB` close. There are still goroutines/workflows outside the root context/errgroup:

- `pkg/hub/server.go:1376`: async provisioning goroutine uses `context.Background()`.
- `pkg/hub/server.go:2296`: bootstrap checkpoint goroutine is launched outside `s.bg`.
- `pkg/hub/server.go:3059`: Daytona bootstrap goroutine uses `context.Background()`.
- `pkg/hub/server.go:4267`: Exedev bootstrap goroutine uses `context.Background()`.
- `pkg/hub/checkpoints.go:558`: pending checkpoint dispatch goroutine uses `context.Background()`.
- `pkg/hub/cron_scheduler.go:219`: cron jobs do not receive the root context, and `runWorkflow` can create claws without cancellation.

During manual SIGTERM validation with a connected claw, the hub also logged:

```text
[checkpoint] bootstrap request for manual-c failed: database is locked (5) (SQLITE_BUSY)
```

That looks like the exact shutdown/SQLite race this phase is meant to eliminate. I would treat this PR as blocked until these remaining goroutines are brought under the root context/errgroup or otherwise explicitly cancelled/waited before DB close.

AnaBerg:
The common execution rules require a CHANGELOG/release note for observable shutdown behavior changes. This PR changes SIGTERM handling, health status during drain, WebSocket closure, and timing, but the diff contains no changelog/release-note entry. Please document the drain sequence and the `shutdown_grace` setting for operators.

PR #443, opened 2026-07-07
feat(hub): remove token from query string, add single-use auth tickets

AnaBerg:
## Motivation

withAuth accepts ?token= as a fallback, so tokens leak into access logs, reverse-proxy logs and URL history. WS/img endpoints move to short-lived single-use tickets.

- `pkg/hub/server.go` `withAuth` fell back to `r.URL.Query().Get("token")` for every `/api/*` route it guards.
- `handleTerminal` (`/api/terminal/{id}`) authenticated **query-param first** (`?token=` before the Authorization header).
- `withConfigMutationAuth` also read the hub token from `?token=`.
- The frontend actively put the long-lived bearer token in URLs: `web/lib/api.ts` built `/api/ws?token=...`, `/api/terminal/{id}?token=...` and `/api/files/view/{id}?path=...&token=...` (the latter used as ``, so the token also lands in browser history/cache and any intermediary logs). `web/hooks/use-hub.ts` even has redaction logic (`describeWsUrl`) to keep the token out of console logs — evidence the leak was a known hazard.

This implements item **0.3 "Token out of the query string"** of the Phase 0 re-architecture spec (`docs/re-arch/phase-0-stop-the-bleeding.md` on the `docs/re-arch` branch).

## Dependencies

Logical notes: independent of the other Phase 0 items; removal of the deprecated `?token=` fallback is scheduled for Phase 2.

## What changed

Backend (`pkg/hub`):

- New `pkg/hub/ticket.go`: in-memory single-use ticket store (32-byte random hex, 30s TTL, redeem-once, opportunistic pruning) plus `POST /api/auth/ticket`. The endpoint authenticates via the `Authorization: Bearer` header **only** — a leaked query token cannot mint tickets. Response: `{"ticket": "...", "expires_in": 30}`.
- `withAuth` now accepts `?ticket=` (redeemed once, resolves to the tenant/GitHub login captured at issue time). The `?token=` fallback still works for one release but logs `deprecated: token in query (path=... remote=...)`.
- `handleTerminal` (`/api/terminal/{id}`) now checks the Authorization header first, then `?ticket=`, and only then the deprecated `?token=` (with the same warning). Previously it checked the query param first.
- `withConfigMutationAuth` keeps its query fallback for the transition but logs the same deprecation warning.
- Grepping the server code for `Query().Get("token")` now only finds: the deprecated fallbacks (all of which warn) and the E2E bridge endpoint, which uses its own dedicated bridge token (out of scope).

Frontend (`web/`):

- `lib/api.ts`: new `fetchAuthTicket()`; `getHubWsUrl()`, `getTerminalWsUrl()` and `getFileViewUrl()` are now async and embed a fresh single-use `?ticket=` instead of the bearer token. If the hub predates `POST /api/auth/ticket` (mixed-version rollout), they fall back to the deprecated `?token=` so old hubs keep working.
- `hooks/use-hub.ts`: `connectWebSocket` awaits the ticketed URL and retries with the existing backoff if the ticket fetch fails. The URL-redaction logic (`describeWsUrl`) stays and now also redacts `?ticket=`.
- `components/terminal.tsx`: `XTerminal` takes only `clawId` and mints its own ticketed URL per connection (tickets are single-use, so a cached URL prop would break reconnects).
- `components/attachment-chip.tsx`: history image URLs are resolved asynchronously; clicking to open the image in a new tab mints a fresh ticket because the one embedded in the `` src has already been consumed.

## Testing

All commands run locally on this branch:

- `go build ./...` — pass.
- `make test` (full `go test -v ./...`) — pass (exit 0). Includes the new `pkg/hub/ticket_test.go`: single-use redemption, expiry, header-only auth on `POST /api/auth/ticket` (query token rejected with 401), `withAuth` accepting `?ticket=` once and rejecting reuse, and the deprecated `?token=` fallback still authenticating.
- `cd web && npm install && npm run build` — pass (compile + TypeScript + static generation all green). The repo has no frontend unit-test script (`package.json` scripts: dev/build/start/lint), so none were run.

Not run: `make e2e` / terminal + file-preview browser E2E (requires ngrok/Daytona credentials not available in this environment).

### DIFF
--- a/pkg/hub/server.go
+++ b/pkg/hub/server.go
@@ -422,8 +422,9 @@ func (s *Server) withAuth(next http.HandlerFunc) http.HandlerFunc {
 		token := strings.TrimPrefix(r.Header.Get("Authorization"), "Bearer ")
 		if token == "" {
 			// Single-use ticket (see ticket.go) — the only supported query-string
-			// auth. Used by WS upgrades and <img src> loads that cannot set headers.
-			if ticket := r.URL.Query().Get("ticket"); ticket != "" {
+			// auth. Only honored on WS upgrades and <img src> loads that cannot
+			// set headers; on any other route the ticket is ignored.
+			if ticket := r.URL.Query().Get("ticket"); ticket != "" && ticketAuthAllowed(r.URL.Path) {
 				tenantID, githubLogin, ok := s.redeemAuthTicket(ticket)
 				if !ok {
 					http.Error(w, "unauthorized", http.StatusUnauthorized)
@@ -6129,21 +6130,23 @@ func remoteWriteFileCommand(dir, name, content string) string {
 func (s *Server) handleTerminal(w http.ResponseWriter, r *http.Request) {
 	// Auth: Authorization header, or a single-use ticket (browsers cannot set
 	// headers on WebSocket upgrades). The raw ?token= fallback is deprecated.
-	var tenantID string
+	// The GitHub login (when present) is preserved so the same per-claw access
+	// checks as the REST/WS paths apply below.
+	var tenantID, ghLogin string
 	if token := strings.TrimPrefix(r.Header.Get("Authorization"), "Bearer "); token != "" {
-		tid, err := s.tenantByToken(token)
-		if err != nil {
-			http.Error(w, "unauthorized", http.StatusUnauthorized)
-			return
-		}
-		tenantID = tid
-	} else if ticket := r.URL.Query().Get("ticket"); ticket != "" {
-		tid, _, ok := s.redeemAuthTicket(ticket)
+		tid, login, ok := s.resolveAuthToken(token)
 		if !ok {
 			http.Error(w, "unauthorized", http.StatusUnauthorized)
 			return
 		}
-		tenantID = tid
+		tenantID, ghLogin = tid, login
+	} else if ticket := r.URL.Query().Get("ticket"); ticket != "" {
+		tid, login, ok := s.redeemAuthTicket(ticket)
+		if !ok {
+			http.Error(w, "unauthorized", http.StatusUnauthorized)
+			return
+		}
+		tenantID, ghLogin = tid, login
 	} else {
 		// Deprecated: raw token in the query string leaks into access logs and
 		// URL history. Kept for one release; removal planned for Phase 2.
@@ -6169,10 +6172,11 @@ func (s *Server) handleTerminal(w http.ResponseWriter, r *http.Request) {
 	var sshHost string
 	var sshPort int
 	var sshUser string
+	var tagsJSON string
 	err := s.db.QueryRow(
-		`SELECT ssh_host, ssh_port, ssh_user FROM claws WHERE id = ? AND tenant_id = ?`,
+		`SELECT ssh_host, ssh_port, ssh_user, COALESCE(tags,'[]') FROM claws WHERE id = ? AND tenant_id = ?`,
 		clawID, tenantID,
-	).Scan(&sshHost, &sshPort, &sshUser)
+	).Scan(&sshHost, &sshPort, &sshUser, &tagsJSON)
 	if err == sql.ErrNoRows {
 		http.Error(w, "not found", http.StatusNotFound)
 		return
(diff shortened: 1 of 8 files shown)


pkg/hub/server.go
```diff
@@ -415,7 +421,29 @@ func (s *Server) withAuth(next http.HandlerFunc) http.HandlerFunc {
 	return func(w http.ResponseWriter, r *http.Request) {
 		token := strings.TrimPrefix(r.Header.Get("Authorization"), "Bearer ")
 		if token == "" {
+			// Single-use ticket (see ticket.go) — the only supported query-string
+			// auth. Used by WS upgrades and <img src> loads that cannot set headers.
+			if ticket := r.URL.Query().Get("ticket"); ticket != "" {
```
AnaBerg: docs/re-arch spec-alignment: Phase 0.3 scopes `?ticket=` to WebSocket and `` resource endpoints, but accepting tickets in the generic `withAuth` wrapper lets a single-use ticket authenticate any `withAuth` REST route, including mutating endpoints that happen to use this wrapper. That widens the authority of a leaked image/WS ticket beyond the spec. Please restrict ticket redemption to `/api/ws` and `/api/files/view/...` here, or move ticket handling onto only the ticket-capable routes.

web/lib/api.ts
```diff
@@ -205,18 +221,32 @@ export async function killClaw(id: string): Promise<void> {
...
+// buildWsUrl authenticates a WebSocket URL with a single-use ticket. Browsers
+// can't set the Authorization header on WS upgrades, so the ticket goes in the
+// query string. Falls back to the deprecated ?token param when the hub
+// predates tickets.
+async function buildWsUrl(path: string): Promise<string> {
+  try {
+    const ticket = await fetchAuthTicket()
+    return `${wsBaseUrl()}${path}?ticket=${encodeURIComponent(ticket)}`
+  } catch {
+    // Deprecated fallback for hubs without POST /api/auth/ticket.
```
AnaBerg, line 253: docs/re-arch spec-alignment: the new UI still falls back to constructing raw `?token=` WebSocket URLs whenever ticket issuance throws. Phase 0.3 keeps server-side `?token=` acceptance only as a one-release transition, while the frontend should request a ticket before opening WS/file-preview URLs so it no longer leaks long-lived tokens into browser/proxy logs. Please fail/retry ticket issuance instead of emitting `?token=`, or at least limit fallback to an explicit legacy-hub detection path rather than every ticket error.

pkg/hub/server.go
```diff
@@ -6092,17 +6125,38 @@ func remoteWriteFileCommand(dir, name, content string) string {
...
+	var tenantID string
+	if token := strings.TrimPrefix(r.Header.Get("Authorization"), "Bearer "); token != "" {
+		tid, err := s.tenantByToken(token)
+		if err != nil {
+			http.Error(w, "unauthorized", http.StatusUnauthorized)
+			return
+		}
+		tenantID = tid
+	} else if ticket := r.URL.Query().Get("ticket"); ticket != "" {
+		tid, _, ok := s.redeemAuthTicket(ticket)
```
AnaBerg: manual review: `redeemAuthTicket` returns the GitHub OAuth login, but the terminal route discards it and only carries `tenantID` into the claw lookup. Because `/api/terminal/` is not wrapped in `withAuth`, an OAuth user who can mint any ticket can connect to the SSH terminal for any claw in the tenant if they know the ID, bypassing the tag filters enforced by list/detail/WS paths (`canViewClaw`/`canModifyClaw`). Please preserve the login from the ticket/header path and apply the same claw access check before dialing SSH; terminal access should probably require `canModifyClaw`, not just tenant ownership.

AnaBerg:
Manual validation found the ticket implementation functional, but I would change direction on the transition policy and remove the `?token=` fallback now.

Positive checks:

- `POST /api/auth/ticket` exists and requires `Authorization: Bearer`.
- Tickets are opaque, 30s TTL, and redeem-once.
- `/api/ws`, `/api/terminal/{id}`, and `/api/files/view/...` can use `?ticket=`.
- Frontend paths mint tickets for WS/file/terminal URLs.
- `make test`, `make test-factory`, focused auth tests, and `npm run build` passed.

The remaining issue is that the spec goal is “token out of the query string”, and keeping transitional `?token=` paths preserves the leakage risk this item is meant to stop. The branch still has `Query().Get("token")` fallbacks in:

- E2E bridge token path.
- `withAuth` fallback.
- config mutation fallback.
- terminal fallback.

Also, the Go client still builds `/api/ws?token=...`, which should be migrated before the fallback is removed.

Recommendation: remove long-lived `?token=` support in this PR and migrate the remaining client/call sites to tickets or headers as appropriate. That makes the implementation match the security intent and the literal grep acceptance, avoids carrying a knowingly unsafe compatibility path, and keeps Phase 0 from shipping with the same token-in-URL class of bug still present.

AnaBerg:
The common execution rules require a CHANGELOG note and a documented transition flag for compatibility-affecting auth changes. This PR introduces tickets and deprecates `?token=`, but the diff contains neither a changelog/release-note entry nor an operator-facing transition flag/documentation. Please add the one-release migration guidance, including affected WebSocket, terminal, and file-view clients.

PR #444, opened 2026-07-07
test(web): vitest unit tests for mappers, attachments, auth-storage, message merge

AnaBerg:
## Motivation

The frontend has zero tests, so the settings split and hook extraction have no safety net. Per the spec, pure-function tests come first, before any refactor: mappers, attachments, auth-storage, then the optimistic message-merge logic.

Concrete evidence from the code:

- `web/package.json` had no test script and no test framework installed at all.
- The optimistic merge in `web/hooks/use-hub.ts` (formerly lines 276–303) is the most fragile logic in the frontend: it reconciles the API timeline with locally cached messages, filters transient (`activity-`/`live-`/`thinking-`) and optimistic (`opt-`) ids, keeps cached history beyond the API fetch window, and drops optimistic messages once the server confirms them by role + content. Any regression here silently duplicates or loses chat messages.
- `web/lib/mappers.ts` already encodes contract-drift workarounds (e.g. `mapApiStatus` maps unknown backend statuses to `offline`, and `starting` to `provisioning`) that were previously undocumented by any test.
- `web/lib/attachments.ts` contains a hand-tuned regex for round-tripping the attachments footer (em-dashes and parentheses in filenames) whose comments describe edge cases nothing verified.

## Dependencies

Note: this is item 3.3 of the phase 3 spec (safety-net-first ordering); items 3.1/3.2 (settings split, hook extraction) should land after this.

## What changed

- **Extracted the merge logic into a pure function** (`refactor` commit): `isTransientMessage` moved from `use-hub.ts` to `web/lib/messages.ts`, plus new `isOptimisticMessage` and `mergeTimelineMessages(existing, incoming)`. `loadMessages` in `use-hub.ts` now calls the extracted function. No behavior change: same filters, same stable sort, API-first ordering on equal timestamps.
- **Vitest setup**: `vitest` + `jsdom` devDependencies, `web/vitest.config.ts` (node environment by default, `@` path alias; DOM-dependent suites opt into jsdom via a per-file docblock), and `npm test` / `npm run test:watch` scripts.
- **Unit tests** (60 tests, 4 files under `web/lib/__tests__/`):
  - `mappers.test.ts` — `parseTagsFromName`, `computeUptime` (fake timers, future `created_at` clamped, unparseable dates), `mapApiStatus` (including unknown statuses degrading to `offline`), `mapApiClaw` (deterministic auto color, override precedence), `mapApiMessage` (activity/activity_summary payload parsing, malformed JSON swallowed).
  - `attachments.test.ts` — `formatBytes`, `buildAttachmentsFooter` (empty/not-ready filtering), `splitAttachmentsFooter` (build/split round-trip, body-less leading-marker form, em-dashes in names, parentheses in filenames, unparseable lines).
  - `auth-storage.test.ts` — sessionStorage token get/set/clear, GitHub-over-hub token precedence, cross-tab BroadcastChannel sync (request/response/update/clear) with a deterministic fake channel, `requestAuthToken` resolution and timeout, OAuth state helpers.
  - `messages.test.ts` — the optimistic merge: id dedupe (API copy wins), cache preserved beyond the API window, transient messages dropped, unconfirmed optimistic messages kept, confirmed ones dropped (role + content match), stable timestamp sort.

Intentionally **not** in this pass (per instructions): the Playwright smoke spec and the CI job in `.depot/workflows/main.yaml`.

## Testing

- `cd web && npm install` — OK.
- `cd web && npm test` — **4 test files, 60 tests, all passing** (vitest 4.1.10).
- `cd web && npm run build` — passes (next build, all routes generated).
- `cd web && npm run lint` — 30 pre-existing problems; verified by stashing this change and re-running lint on `main`: identical 30 problems, so this PR introduces none.
- No Go changes, so `go build` / `make test` were not run.

### DIFF
--- a/web/package.json
+++ b/web/package.json
@@ -6,7 +6,9 @@
     "dev": "next dev",
     "build": "next build",
     "start": "next start",
-    "lint": "eslint ."
+    "lint": "eslint .",
+    "test": "vitest run",
+    "test:watch": "vitest"
   },
   "dependencies": {
     "@dnd-kit/core": "^6.3.1",
@@ -75,10 +77,12 @@
     "@types/react-dom": "^19",
     "eslint": "^9.39.4",
     "eslint-config-next": "^16.2.0",
+    "jsdom": "^29.1.1",
     "postcss": "^8.5",
     "tailwindcss": "^4.2.0",
     "tw-animate-css": "1.3.3",
-    "typescript": "5.7.3"
+    "typescript": "5.7.3",
+    "vitest": "^4.1.10"
   },
   "optionalDependencies": {
     "@tailwindcss/oxide-linux-arm64-gnu": "4.2.0",
--- a/web/vitest.config.ts
+++ b/web/vitest.config.ts
@@ -1,0 +1,16 @@
+import path from "node:path"
+import { defineConfig } from "vitest/config"
+
+export default defineConfig({
+  resolve: {
+    alias: {
+      "@": path.resolve(__dirname, "."),
+    },
+  },
+  test: {
+    // Pure-function tests run in node; DOM-dependent suites opt into jsdom
+    // with a `// @vitest-environment jsdom` docblock.
+    environment: "node",
+    include: ["**/__tests__/**/*.test.ts"],
+  },
+})
--- a/web/vitest.config.ts
+++ b/web/vitest.config.ts
@@ -11,6 +11,6 @@ export default defineConfig({
     // Pure-function tests run in node; DOM-dependent suites opt into jsdom
     // with a `// @vitest-environment jsdom` docblock.
     environment: "node",
-    include: ["**/__tests__/**/*.test.ts"],
+    include: ["**/__tests__/**/*.{test,spec}.{ts,tsx}"],
   },
 })
--- a/.depot/workflows/test.yaml
+++ b/.depot/workflows/test.yaml
@@ -75,6 +75,10 @@ jobs:
         run: npm install
         working-directory: web
 
+      - name: Unit tests
+        run: npm test
+        working-directory: web
+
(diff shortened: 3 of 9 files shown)


web/package.json
```diff
@@ -6,7 +6,9 @@
     "dev": "next dev",
     "build": "next build",
     "start": "next start",
-    "lint": "eslint ."
+    "lint": "eslint .",
+    "test": "vitest run",
```
AnaBerg, line 10: docs/re-arch spec-alignment: Phase 3.3 calls for the new Vitest unit suite to be a CI gate, but this PR only adds the local `npm test` script and leaves `.depot/workflows/test.yaml` running lint/build without `npm test` in `web`. Please add a PR workflow step in the web job, after `npm install`, that runs `npm test` with `working-directory: web` so these frontend unit tests cannot regress silently.

web/vitest.config.ts
```diff
@@ -0,0 +1,16 @@
...
+  resolve: {
+    alias: {
+      "@": path.resolve(__dirname, "."),
+    },
+  },
+  test: {
+    // Pure-function tests run in node; DOM-dependent suites opt into jsdom
+    // with a `// @vitest-environment jsdom` docblock.
+    environment: "node",
+    include: ["**/__tests__/**/*.test.ts"],
```
AnaBerg: manual review: This custom include only discovers `.test.ts` files, so any React/component or hook tests added as `.test.tsx` will be silently skipped by `npm test`. Phase 3.3 specifically calls out React/testing-library coverage for the message-cache path, so the harness should include TSX tests as well, e.g. `**/__tests__/**/*.{test,spec}.{ts,tsx}` or rely on Vitest's default include patterns.

PR #445, opened 2026-07-07
feat(hub): single typed config loader with SIGHUP hot reload

AnaBerg:
Implements item **3.6 Config consolidation** from the re-architecture plan (`docs/re-arch/phase-3-frontend-and-hardening.md` on `docs/re-arch`).

## Motivation

Two config systems coexist (Viper for the CLI, manual YAML for the hub) and hub.yaml has no hot reload. A single typed loader in pkg/config/hub.go with documented precedence and SIGHUP reload for the safe subset consolidates it.

- `cmd/root.go` configures Viper (`SetEnvPrefix("ELASTICCLAW")`, `AutomaticEnv`, `~/.elasticclaw/config.yaml`) for the CLI, while `cmd/hub.go` loaded `hub.yaml` through the manual `config.LoadHubConfig()` — two unrelated resolution paths in the same binary.
- The hub's process settings were flag-only: `--addr` and `--db` had no env or hub.yaml equivalents, and there was no log level at all, so no documented precedence existed for the server.
- Any hub.yaml edit (CORS, branding, integrations) required a full restart; `corsMiddleware` in `pkg/hub/server.go` hard-coded `Access-Control-Allow-Origin: *`.

## Dependencies

Logical notes: this deliberately does **not** touch the Phase 3.4 secrets envelope or 3.5 JWT work; the settings API's existing `s.hubCfg = &updatedCfg` swap pattern is reused for the reload path.

## What changed

- **New hub.yaml fields** (`pkg/types/template.go`): `listen_addr`, `db_path` (restart required), `log_level`, `allowed_origins` (hot-reloadable). All optional; omitted fields keep the previous behavior exactly.
- **Consolidated loader** (`pkg/config/hub.go`): `LoadHubBootConfig` resolves addr / db path / log level / allowed origins with the documented precedence **flag > env (`ELASTICCLAW_ADDR`, `ELASTICCLAW_DB`, `ELASTICCLAW_LOG_LEVEL`, `ELASTICCLAW_ALLOWED_ORIGINS`) > hub.yaml > default**, and validates them at boot (log level enum, address shape). Viper stays CLI-only — the hub never touches it.
- **SIGHUP hot reload**: `cmd/hub.go` installs a SIGHUP handler; `Server.ApplyHotReload` (backed by `config.ApplyHotReload`) applies the safe subset — `log_level`, `allowed_origins`, `branding`, `integrations` — under the server mutex. Changes to `listen_addr`/`db_path` on disk are rejected with a clear log stating a restart is required. Flag/env-pinned log levels are re-resolved on reload so a yaml edit cannot override `--log-level` or the env var.
- **Live log level**: the standard `log` package is routed through `slog` with a `slog.LevelVar`, so `kill -HUP` after a `log_level` change takes effect immediately (e.g. `error` silences the hub's info logging). Side effect: hub log lines now carry the slog text format (`time=... level=... msg=...`).
- **CORS honors `allowed_origins`**: `corsMiddleware` is now a `Server` method; an empty list keeps the historical `*` behavior, otherwise only listed origins are echoed (with `Vary: Origin`), read per request so reloads apply instantly.
- **Tests** (`pkg/config/hub_test.go`): precedence matrix (default / yaml / env / flag) for addr, db path, and log level; env-beats-yaml for allowed origins; boot validation failures; `ParseLogLevel`; hot-reload safe-subset application, restart-field rejection, and input non-mutation.

## Testing

- `go build ./...` — pass.
- `go test ./...` (same packages as `make test`, without `-v`) — all packages pass, no failures.
- `go test ./pkg/config/` — pass, including the new precedence matrix and hot-reload tests.
- Manual acceptance run of the built binary:
  - started `elasticclaw hub --addr :18099` with a temp `hub.yaml` (`log_level: info`), edited it to `error`, sent `kill -HUP` → log shows `SIGHUP reload: applied log_level` and `log level INFO -> ERROR`; a second reload's info-level line was correctly suppressed at `error`, and a later `debug` edit + HUP switched to `DEBUG` — no restart.
  - CORS: default responses carry `Access-Control-Allow-Origin: *`; after appending `allowed_origins: [https://good.example.com]` and `kill -HUP`, a matching `Origin` is echoed with `Vary: Origin` and a non-matching origin gets no `Access-Control-Allow-Origin` header.

### DIFF
--- a/cmd/hub.go
+++ b/cmd/hub.go
@@ -2,9 +2,13 @@ package cmd
 
 import (
 	"fmt"
+	"log"
+	"log/slog"
 	"os"
+	"os/signal"
 	"path/filepath"
 	"strings"
+	"syscall"
 
 	"github.com/elasticclaw/elasticclaw/pkg/config"
 	"github.com/elasticclaw/elasticclaw/pkg/hub"
@@ -18,6 +22,7 @@ var (
 	hubToken      string
 	hubClawToken  string
 	hubUIPassword string
+	hubLogLevel   string
 	hubNoWebUI    bool
 )
 
@@ -38,8 +43,9 @@ The hub stores data in a SQLite database (~/.elasticclaw/hub.db by default).
 func init() {
 	rootCmd.AddCommand(hubCmd)
 
-	hubCmd.Flags().StringVar(&hubAddr, "addr", ":8080", "address to listen on")
+	hubCmd.Flags().StringVar(&hubAddr, "addr", "", "address to listen on (default: "+config.DefaultHubAddr+")")
 	hubCmd.Flags().StringVar(&hubDBPath, "db", "", "path to SQLite database (default: ~/.elasticclaw/hub.db)")
+	hubCmd.Flags().StringVar(&hubLogLevel, "log-level", "", "log level: debug, info, warn, error (default: info; hot-reloadable via SIGHUP)")
 	hubCmd.Flags().StringVar(&hubToken, "token", "", "create/update default tenant with this user token")
 	hubCmd.Flags().StringVar(&hubClawToken, "claw-token", "", "agent authentication token (required with --token)")
 	hubCmd.Flags().StringVar(&hubUIPassword, "ui-token", "", "web UI login password (default: value from hub.yaml ui_password, or 'admin')")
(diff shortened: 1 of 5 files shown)


AnaBerg:
Manual validation against `docs/re-arch` found a CORS default mismatch.

Phase 0.4 requires an omitted `allowed_origins` setting to default to the hub's own browser-facing origin. This branch instead preserves the historical wildcard: with no `allowed_origins` configured, a request with `Origin: https://evil.example` returned:

```text
Access-Control-Allow-Origin: *
```

That leaves the insecure default in place and conflicts with the re-architecture contract this config work is expected to consolidate. Please derive the default allowlist from the browser-facing hub URL (and keep `Vary: Origin`) rather than returning `*`.

AnaBerg:
This branch defines an empty `allowed_origins` list as `Access-Control-Allow-Origin: *` and also accepts an explicit `*`. That contradicts the re-architecture spec: Phase 0.4 requires the default to be the hub's own origin and unlisted origins to receive no CORS header. Because this PR is based on `main`, merging it independently would preserve/reintroduce the permissive policy instead of consolidating the hardened configuration. Please align the loader and `resolveCORSOrigin` with the restricted-CORS behavior (own-origin default, no wildcard fallback).

PR #446, opened 2026-07-07
feat(hub): versioned embedded SQL migrations with schema_migrations table

AnaBerg:
Implements item **1.4 Versioned migrations** of the Phase 1 re-architecture spec (`docs/re-arch/phase-1-contract-and-observability.md` on `docs/re-arch`).

## Motivation

Migrations today are unversioned idempotent ALTER TABLEs; there is no record of applied schema versions and no transactional boundary. Numbered embedded migrations with baseline detection for existing installs fix this.

Concretely, `pkg/hub/db.go` ran the same ~40 `ALTER TABLE ... ADD COLUMN` statements (tolerating "duplicate column name") plus 5 guarded data migrations on **every boot**, on top of a ~440-line `CREATE TABLE IF NOT EXISTS` block — all outside any transaction, so a mid-migration failure (full disk, killed process) could leave the schema half-applied with nothing recording what state the database was in. Phase 0.5 (#440) stopped swallowing the errors, but the schema still had no version.

## Dependencies

This PR is STACKED on branch `feature/migration-error-handling`. Depends on #440 — review and merge that first.

## What changed

- **New `pkg/hub/store` package** owning schema migrations:
  - `migrations/0001_init.sql` — the current consolidated schema, embedded via `//go:embed` (the existing `CREATE TABLE IF NOT EXISTS` block; every historical ALTER column was already collapsed into those CREATEs).
  - `migrate.go` — minimal in-house runner (per the spec's preference over `golang-migrate`; ~110 LOC of code): loads `NNNN_name.sql` files, validates numbering (contiguous, no duplicates, starts at 1), records applied versions in a new `schema_migrations` table, and applies each pending migration **inside its own transaction** — failure rolls back and aborts boot. A database reporting a version newer than the binary is rejected with an explicit "downgrades are not supported" error (documented in the package doc; roll back by restoring `hub.db` from backup).
  - `legacy.go` — existing-install detection: if `claws` exists but `schema_migrations` does not, the runner performs one final transactional run of the legacy idempotent catch-up (base `IF NOT EXISTS` schema + tolerated ALTERs + guarded data migrations, moved verbatim from `db.go` and frozen) and then registers the baseline as version 1 without executing `0001_init.sql` as a migration. The catch-up step goes slightly beyond the spec's "register the baseline without executing it": it preserves today's behavior where a database of *any* previous vintage (not just the immediately prior release) gains its missing columns on upgrade.
- **`pkg/hub/db.go`** shrinks to `openDB` delegating to `store.Migrate`; the inline `migrate()`, `execIgnoreDuplicate` and `summarizeStmt` moved to the store package.
- **Committed old-DB fixture** `pkg/hub/store/testdata/hub-legacy.db` (generator committed alongside): a mid-vintage database with old `claws`/`messages` shapes and real rows. The upgrade test asserts new columns and tables appear, data survives, the shortcut-ID data migration runs, a shortcut factory trigger is backfilled, and the DB is baselined at version 1.

New columns/tables from here on go into `0002_*.sql` etc.; the legacy statement lists are frozen.

## Testing

- `go build ./...` — pass.
- `go test ./...` (same package set as `make test`, without `-v`) — pass, exit 0.
- `go test ./pkg/hub/store/... -v` — 4/4 pass: `TestMigrateFreshInstall`, `TestMigrateLegacyFixtureUpgrade` (committed fixture upgrade path), `TestMigrateRejectsNewerDatabase`, `TestLoadMigrationsAreWellFormed`.
- `go test ./pkg/hub/ -run 'TestMigrate|TestClaws' -v` — pass, including the pre-existing `TestMigrateIsIdempotent`, `TestMigrateFailsOnReadOnlyDatabase` and both `TestClawsTriggerActorJSONDefaultIsValidJSON` subtests (the "migrated schema" subtest now exercises the legacy-baseline path).


AnaBerg:
The legacy-database detection is not atomic with baselining. `schema_migrations` is created and committed before `legacyBaseline` runs. If the baseline transaction then fails (lock, disk-full, etc.), the next boot sees `schema_migrations` and skips the legacy path; it runs `0001_init.sql`, whose `CREATE TABLE IF NOT EXISTS` statements do not add missing columns to existing tables, and then records version 1. That can permanently mark an incomplete legacy schema as current. Please create the version table and record the legacy baseline in the same transaction, or make detection also treat an empty version table plus existing legacy tables as needing `legacyBaseline`.

PR #447, opened 2026-07-07
test(provider): in-memory mock provider and handler unit tests

AnaBerg:
## Motivation

Handler tests today need Daytona/Docker. An in-memory Provider implementation with programmable states and injectable failures enables direct unit tests for the main handlers and a BenchmarkClawWSMessage baseline.

Concrete evidence from the code:

- Every concrete `provider.Provider` implementation (`pkg/provider/daytona`, `docker`, `exedev`, `lambdamicrovms`, `replicated`) talks to an external service or daemon, so nothing implementing the interface could be used in unit tests.
- `pkg/hub/server.go` is ~6,560 lines and, before this PR, the main claw/message handlers (`handleCreateClaw`, `handleMessages`, `handleClaws`) had no direct request/response unit tests for their core success paths, and neither did `handleSettings` GET/PATCH (`pkg/hub/settings.go`) beyond `getSettings` edge cases; the Linear webhook tests (`pkg/hub/linear_webhook_test.go`) covered workflow behavior but not the handler's method/signature/payload validation.
- The claw WebSocket hot path (decode -> persist -> broadcast in `handleClawWS`) had no benchmark, so there was no baseline to measure future optimizations against.

This implements item 3.7 (Backend testability) of the re-architecture plan.

## Dependencies

Note: Phase 2's handler extraction has not landed on main yet, so the unit tests target the handlers where they live today (`pkg/hub/server.go`, `settings.go`, `linear.go`). They exercise handlers directly via `httptest`, so they will survive a later extraction. Also note the create-claw path dispatches provisioning by provider name (`provisionDaytona`/`provisionDocker`/...) rather than through the `provider.Provider` interface, so the mock cannot be wired into that path yet; the handler tests use the in-memory SQLite store from `NewTestServerWithConfig`, and the mock is exercised through its own suite and the provider registry.

## What changed

- **`pkg/provider/mock`** (new): in-memory, thread-safe implementation of `provider.Provider` with:
  - programmable instance states (`SetStatus`) and programmable `Exec`/`Connect` results;
  - injectable failures per operation, persistent (`Fail`) or one-shot (`FailOnce`);
  - context-cancellation awareness and call recording (`Calls`) for assertions;
  - full test suite: lifecycle, failure injection, cancellation, registry integration, concurrency.
- **`pkg/hub/handler_unit_test.go`** (new): direct unit tests for 5 main handlers against the in-memory store:
  - create claw: validation (400/422) and success (202 + persisted row);
  - send message: persistence + response shape, empty content rejected;
  - list claws: stale `connected` rows are normalized to `offline`, `provisioning`/`error` preserved;
  - settings GET: raw LLM API keys are never leaked (masked as `keySet`);
  - settings PATCH: LLM key upsert applied in memory and persisted to hub.yaml, 405 for other methods;
  - Linear webhook: 405 on GET, 401 on missing/wrong HMAC signature, 200 on valid non-Issue events, 400 on invalid JSON with a valid signature.
- **`pkg/hub/claw_ws_bench_test.go`** (new): `BenchmarkClawWSMessage` exercises the real hot path end to end over WebSockets (claw sends a `message` frame -> hub decodes -> persists to SQLite -> broadcasts to a connected user WS). Baseline on a dev laptop (VirtualApple @ 2.50GHz): ~262µs/op, ~10.2KB/op, 251 allocs/op.

No production code was changed — this PR is tests only.

## Testing

- `go build ./...` — pass.
- `go test ./pkg/provider/mock/...` — pass (spec acceptance).
- `go test ./pkg/hub/ -run 'TestHandleCreateClaw|TestHandleMessagesPost|TestHandleClawsListMarks|TestSettingsGetMasks|TestSettingsPatchUpserts|TestHandleLinearWebhook' -v` — all pass.
- `go test ./pkg/hub/ -run '^$' -bench BenchmarkClawWSMessage -benchmem -benchtime 200x` — pass (262244 ns/op, 10249 B/op, 251 allocs/op).
- `go test ./...` (equivalent of `make test` without `-v`) — full suite passes, no failures.

### DIFF
--- a/pkg/hub/claw_ws_bench_test.go
+++ b/pkg/hub/claw_ws_bench_test.go
@@ -1,0 +1,107 @@
+package hub
+
+// BenchmarkClawWSMessage measures the claw WebSocket message hot path
+// (decode -> persist -> broadcast) end to end: a registered claw connection
+// sends a complete "message" frame, the hub decodes it, persists it to the
+// in-memory SQLite store, and broadcasts it to a connected user WebSocket.
+// This is the baseline for future optimizations (re-architecture plan 3.7).
+//
+// Run with:
+//
+//	go test ./pkg/hub/ -run '^$' -bench BenchmarkClawWSMessage -benchmem
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"io"
+	"log"
+	"net/http/httptest"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/elasticclaw/elasticclaw/pkg/types"
+	"nhooyr.io/websocket"
+	"nhooyr.io/websocket/wsjson"
+)
+
+func BenchmarkClawWSMessage(b *testing.B) {
+	b.Setenv("ELASTICCLAW_HUB_CONFIG", b.TempDir()+"/hub.yaml")
+	// Silence per-message hub logging during the benchmark.
+	prevOut := log.Writer()
+	log.SetOutput(io.Discard)
+	b.Cleanup(func() { log.SetOutput(prevOut) })
+
+	s, _ := NewTestServerWithConfig(b, &types.HubConfig{ClawToken: "claw-token"}, "", "", "")
+	ts := httptest.NewServer(s.Handler())
+	b.Cleanup(ts.Close)
+
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Minute)
+	defer cancel()
+	wsBase := "ws" + strings.TrimPrefix(ts.URL, "http")
+
+	// User connection that receives the broadcast.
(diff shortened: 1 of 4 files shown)


AnaBerg:
The Phase 3.7 requirement is specifically to test extracted handlers with the in-memory provider mock plus the in-memory store. This PR creates both pieces but does not connect them: the PR description acknowledges that the create-claw path still dispatches directly by provider name, and none of the handler tests inject `pkg/provider/mock`. As a result provider success/failure behavior at the handler boundary remains untested, which is the main testability gap this item was intended to close. Please add a provider factory/registry seam to the handler and use the mock in at least the create-claw success/failure tests.

PR #450, opened 2026-07-08
feat(hub): encrypt secrets at rest in hub.yaml (AES-256-GCM envelope)

AnaBerg:
Implements item **3.4 Secrets at rest** of the Phase 3 re-architecture spec (`docs/re-arch/phase-3-frontend-and-hardening.md` on `docs/re-arch`).

## Motivation

hub.yaml stores the GitHub App private key, tracker tokens and passwords in plaintext. Sensitive fields get an `enc:v1:` AES-256-GCM envelope with a master key from env or `~/.elasticclaw/master.key`, with transparent migration on save.

- `types.HubConfig` (`pkg/types/template.go`) carries `Token`, `ClawToken`, `UIPassword`, `Auth.SessionSecret`, `GitHubApps[].PrivateKeyPEM` (PEM-encoded RSA key pasted directly in yaml), `LLMKeys[].APIKey`, provider credentials (`ProviderConfig.APIKey`/`AccessToken`/`Token`), the `Secrets` map, and Linear/Shortcut/Jira/GitHub-Issues integration tokens and webhook secrets — all serialized verbatim by `yaml.Marshal` in `pkg/config/hub.go` and written to disk.
- The installer (`pkg/install/scripts.go`, `HubConfig()`) emitted `token`, `claw_token` and `ui_password` in cleartext in the generated hub.yaml on every fresh install.

## Dependencies

Note: spec item 3.5 (golang-jwt sessions) derives its signing key from this master key via HKDF, so it should branch from (or follow) this PR.

## What changed

- **New `pkg/secrets` package**
  - `enc:v1:<base64(nonce || ciphertext)>` envelope, AES-256-GCM under a 32-byte master key.
  - Master key resolution: `ELASTICCLAW_MASTER_KEY` (base64) first, then a `master.key` file (created on first boot, 0600) — `~/.elasticclaw/master.key`, or next to the config when `ELASTICCLAW_HUB_CONFIG` is set.
  - Reflective `EncryptFields`/`DecryptFields` walk structs for fields tagged `secret:"true"` (string and `map[string]string`); idempotent, plaintext passthrough on decrypt for transparent migration.
- **`pkg/types/template.go`**: `secret:"true"` tags on all sensitive fields (hub tokens, UI password, session secret, GitHub App private key, LLM API keys, provider credentials, tracker tokens, webhook secrets, the `secrets` map). Non-sensitive fields stay plaintext, so hub.yaml remains readable/editable.
- **`pkg/config/hub.go`**: `SaveHubConfig` encrypts tagged fields on a deep copy (the caller's in-memory config is never mutated); `LoadHubConfig` transparently decrypts and still accepts plaintext values, which are re-written encrypted on the next save. Loading an encrypted config without the master key fails with a message pointing at the runbook.
- **`cmd/hub_encrypt_secrets.go`**: new `elasticclaw hub encrypt-secrets` command forces the full migration immediately.
- **`cmd/hub.go`**: the hub ensures the master key exists on boot (first-boot generation, 0600).
- **Installer** (`pkg/install/scripts.go`, `cmd/install.go`): generates a master key, writes it to a 0600 `EnvironmentFile` at `/etc/elasticclaw/master.env`, references it from the systemd unit (`EnvironmentFile=-/etc/elasticclaw/master.env`), and writes `token`/`claw_token`/`ui_password` in the generated hub.yaml as `enc:v1:` envelopes — a fresh install's hub.yaml contains no cleartext secret.
- **`docs/secrets-at-rest.md`**: format, key resolution, migration, and the lost-master-key runbook (re-enter secrets).

## Testing

All run locally on this branch:

- `go build ./...` — pass.
- `go test ./...` (same package set as `make test`, without `-v`) — all packages pass, no failures.

New tests:

- `pkg/secrets/secrets_test.go`: envelope roundtrip, wrong-key failure, plaintext passthrough, tagged-field walking, master key load/ensure.
- `pkg/config/hub_test.go`: secrets encrypted at rest (raw file contains no plaintext secret, non-secret fields stay readable), load decrypts transparently, plaintext legacy config accepted and migrated on save, missing master key fails loudly, save does not mutate the caller's config.
- `pkg/install/scripts_test.go`: generated hub.yaml has no cleartext secrets when a master key is set, master.env content/script (umask 077 + chmod 600, shellcheck-clean), systemd unit references the EnvironmentFile.

### DIFF
--- a/cmd/hub.go
+++ b/cmd/hub.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/elasticclaw/elasticclaw/pkg/config"
 	"github.com/elasticclaw/elasticclaw/pkg/hub"
+	"github.com/elasticclaw/elasticclaw/pkg/secrets"
 	"github.com/elasticclaw/elasticclaw/pkg/types"
 	"github.com/spf13/cobra"
 )
@@ -66,6 +67,13 @@ func runHub(cmd *cobra.Command, args []string) error {
 	dbPath := hubDBPath
 	if dbPath == "" {
 		dbPath = filepath.Join(dir, "hub.db")
+	}
+
+	// Ensure the secrets master key exists (created on first boot, 0600).
+	// It is required to decrypt enc:v1: values in hub.yaml and to encrypt
+	// sensitive fields on save.
+	if _, err := secrets.EnsureMasterKey(); err != nil {
+		return fmt.Errorf("failed to initialize secrets master key: %w", err)
 	}
 
 	hubCfg, err := config.LoadHubConfig()
--- a/cmd/hub_encrypt_secrets.go
+++ b/cmd/hub_encrypt_secrets.go
@@ -1,0 +1,41 @@
+package cmd
+
+import (
+	"fmt"
+
+	"github.com/elasticclaw/elasticclaw/pkg/config"
+	"github.com/spf13/cobra"
+)
+
+var hubEncryptSecretsCmd = &cobra.Command{
+	Use:   "encrypt-secrets",
+	Short: "Encrypt all sensitive fields in hub.yaml at rest",
+	Long: `Rewrite hub.yaml so every sensitive field (tokens, passwords, private keys,
+webhook secrets) is stored as an enc:v1: AES-256-GCM envelope.
+
+The master key is read from the ELASTICCLAW_MASTER_KEY environment variable
+(32 bytes, base64) or from the master.key file next to the hub data
+(~/.elasticclaw/master.key by default); a new key is generated on first use.
+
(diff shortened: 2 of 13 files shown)


AnaBerg:
The installer fails open on encryption errors: `encryptOrPlaintext` silently returns the original secret. The Phase 3.4 acceptance criterion is that a fresh install's `hub.yaml` contains no cleartext secret; a nonce-generation/encryption failure must therefore abort installation, not write credentials in plaintext. Please return and propagate the error from `HubConfig` generation (or otherwise make plaintext fallback impossible).

PR #451, opened 2026-07-08
feat(api): OpenAPI spec + Go/TS codegen for UI endpoints, fix status drift

AnaBerg:
## Motivation

The Go<->TypeScript contract is mirrored by hand across ~90 endpoints and has already drifted: `InstanceStatus` in Go has `starting`/`deleted` which the TS `ClawStatus` union does not know. Spec-first OpenAPI with oapi-codegen (Go) and openapi-typescript (TS) makes the contract generated, and the generated enum forces the UI to handle every status.

Concrete drift found while writing the spec against the actual handlers:

- The hub writes `starting` and `deleted` into `claws.status` (e.g. `pkg/hub/github_issues.go:581` sets `status='deleted'`; the docker/daytona/lambdamicrovms providers all report `types.StatusStarting`), but the hand-written `ApiClaw.status` union in `web/lib/types.ts` listed neither, so `mapApiStatus` silently rendered them through its `default: return "offline"` branch — a "starting" claw showed as offline instead of spinning, and deleted claws could render in the list.
- `CreateClawRequest` in `web/lib/types.ts` declared a `template` field, but the Go create handler decodes `template_name` (`pkg/types/template.go:780`). The UI was sending a field the server ignores.

## Dependencies

Implements item 1.1 (batch 1 only) + 1.2 of `docs/re-arch/phase-1-contract-and-observability.md` (spec lives on the `docs/re-arch` branch). Batches 2/3 (integration webhooks, internal claw<->hub endpoints), CI wiring of `gen-check`, handler adoption of the generated Go server interfaces, and the `/api/docs` Redoc page are follow-ups per the incremental-scope note in the spec.

## What changed

- **`api/openapi.yaml`** — spec-first contract for the batch-1 UI endpoints: `/api/claws*`, `/api/messages*`, `/api/auth/*`, `/api/settings*`, `/api/analytics/*`, `/api/templates*` (33 paths). Schemas were written from the current Go handler behavior, including the full 8-value `ClawStatus` enum (`pending`, `provisioning`, `starting`, `connected`, `idle`, `offline`, `error`, `deleted`).
- **Go codegen** — `api/gen/api.gen.go` via `oapi-codegen` (std-http mode, ServeMux-compatible server interface + types), configured in `api/oapi-codegen.yaml` and pinned with a `go.mod` `tool` directive. Existing handlers are not migrated yet (gradual adoption per spec).
- **TS codegen** — `web/lib/gen/api.d.ts` via `openapi-typescript` (`npm run gen`).
- **`make gen` / `make gen-check`** — regenerate both sides / fail when the committed artifacts are stale. Not wired into CI in this PR (explicitly out of scope for this pass).
- **`web/lib/types.ts`** — no longer hand-defines any batch-1 wire type; `ApiClaw`, `ApiMessage`, `CreateClawRequest`, `ApiClawStatus`, and the `TaskRun*` analytics types are derived from the generated `components["schemas"]`. Presentation types (`Claw`, `Message`, ...) stay hand-written.
- **Status drift fix (1.2)** — `mapApiStatus` in `web/lib/mappers.ts` is now an exhaustive switch over the generated enum: `starting` renders like `provisioning` (spinner); `deleted` is filtered out of the claw list (`isDeletedClaw`, applied in `useHub.mergeClaws`). New `assertNever` helper in `web/lib/utils.ts` makes any future status a compile error instead of a silent fallthrough, backed by a compile-time completeness check on `API_CLAW_STATUSES`.
- **`web/lib/mappers.test.ts`** — vitest mapper test covering every value of the generated enum (vitest added as a devDependency with an `npm test` script).
- **Bug fix surfaced by the generated types** — `useHub.createClaw` now sends `template_name` (what the Go handler actually decodes) instead of `template`.

## Testing

- `go build ./...` — pass.
- `go test -count=1 ./...` — pass (full suite, uncached).
- `cd web && npm install && npm run build` — pass (Next.js build including type-check).
- `cd web && npx tsc --noEmit` — pass.
- `cd web && npm run test` — pass (12 tests: exhaustive status-mapping table over the generated enum, starting==provisioning presentation, deleted filtering, assertNever throw).
- `make gen` then `git status` — no diff (generated artifacts are fresh); `make gen-check` — pass.


AnaBerg:
There are two unmet Phase 1.1 acceptance criteria in the current head:

- `make gen-check` exists, but no Depot/GitHub workflow invokes it, so stale generated Go/TypeScript artifacts are not rejected in CI.
- The development-only browsable `/api/docs` endpoint (Redoc/Scalar) is absent.

The PR description calls both follow-ups, but `docs/re-arch/phase-1-contract-and-observability.md` lists them as acceptance requirements for the OpenAPI item. Please either include both in this PR or explicitly scope this PR as an incomplete first slice rather than completion of item 1.1.

AnaBerg:
This PR does not meet several explicit Phase 1.1 acceptance items: `make gen-check` is defined but is not invoked by any CI workflow; no `/api/docs` development-only route or Redoc/Scalar artifact is added; and the batch-1 contract is already incomplete when combined with the auth-ticket work because `/api/auth/ticket` is absent. Please wire `gen-check` into CI, add the dev-only browsable docs endpoint, and include the complete `/api/auth/*` surface so the YAML can actually remain the single source of truth.

PR #452, opened 2026-07-08
feat(hub): standard JWT sessions via golang-jwt with HKDF-derived key

AnaBerg:
## Motivation

`pkg/hub/auth_github.go` hand-rolls its own signed session token format — `base64url(json payload).base64url(hmac-sha256)` with manual `hmac.New(sha256.New, ...)` signing and verification (`signGitHubSession`/`verifyGitHubSession`) — even though `github.com/golang-jwt/jwt/v5 v5.3.1` is already in `go.mod` and used elsewhere in the hub (`pkg/hub/github.go` signs GitHub App JWTs with it). Worse, the HMAC key defaulted to the raw `hubCfg.Token` (the shared hub password) when no `session_secret` was configured, so session-forging key material was the same string users type to log in.

This PR implements re-arch item 3.5: sessions become standard HS256 JWTs (claims `sub`, `exp` 7d, `iat`, plus custom `login`/`name`/`avatar`), signed with a key derived from the 3.4 master key via HKDF-SHA256 with a purpose-bound info string (`elasticclaw/hub session-jwt v1`) — no longer the raw `hubCfg.Token`. Verification is dual for one release (JWT first, then the legacy format), so existing sessions survive the upgrade.

## Dependencies

This PR is STACKED on branch `feature/secrets-encryption` (item 3.4 — the master key this derives from). Depends on #450 — review and merge that first.

## What changed

- **`pkg/hub/auth_session.go` (new)** — JWT session module:
  - `deriveSessionJWTKey`: HKDF-SHA256 (stdlib `crypto/hkdf`, Go >= 1.24) from the 32-byte master key, domain-separated info string, 32-byte output.
  - `Server.sessionSigningKey()`: lazily loads the master key (`secrets.LoadMasterKey`, the same source `cmd/hub.go` ensures at boot) and caches the derived key via `sync.Once`; returns nil when no master key exists.
  - `Server.signSession()`: issues HS256 JWTs; falls back to the legacy HMAC format on servers without a master key; errors instead of signing with an empty key when neither exists.
  - `Server.verifySession()`: parses/validates the JWT (`WithValidMethods([HS256])`, `WithExpirationRequired`), then falls back to legacy `verifyGitHubSession` — the one-release transition window, marked with a `TODO(next release)` to drop it.
- **`pkg/hub/server.go`** — the four auth sites (admin middleware, `resolveAuthToken`, web session middleware, `withWebAdminAuth`) now call `s.verifySession(token)` instead of `verifyGitHubSession(s.webSessionSecret(), token)`. Added `sessionJWTOnce`/`sessionJWTKey` fields to `Server`.
- **`pkg/hub/auth_github.go`** — `handleGitHubOAuthExchange` issues tokens via `s.signSession(...)`; legacy `signGitHubSession`/`verifyGitHubSession`/`webSessionSecret` are kept but marked `Deprecated` (verification-only transition + no-master-key fallback).
- **`pkg/hub/auth_session_test.go` (new)** — issuance/validation/expiry tests: JWT round trip; standard claims (`sub`/`iat`/`exp`); expired, wrong-key, tampered, and `alg=none` tokens rejected; legacy tokens still verifying while a JWT key is present; deterministic HKDF derivation (sessions survive hub restarts) that never equals the raw master key; and the no-master-key legacy signing fallback.

No frontend changes: the token remains an opaque string to the web UI.

## Testing

- `go build ./...` — pass.
- `go test ./pkg/hub/ -run 'TestSession' -v` — all 9 new tests pass.
- `make test` (full `go test -v ./...`) — exit 0, no failures; the existing session tests in `pkg/hub/server_test.go` (which sign legacy-format tokens directly) still pass thanks to dual verification.

### DIFF
--- a/pkg/hub/auth_github.go
+++ b/pkg/hub/auth_github.go
@@ -29,8 +29,11 @@ type githubSessionPayload struct {
 	Exp       int64  `json:"exp"`
 }
 
-// signGitHubSession creates a signed session token for a GitHub user.
+// signGitHubSession creates a legacy signed session token for a GitHub user.
 // Format: base64url(payload).base64url(hmac-sha256)
+//
+// Deprecated: new sessions are standard JWTs (see auth_session.go). This is
+// kept only as a fallback for servers without a master key.
 func signGitHubSession(secret, login, name, avatarURL string) (string, error) {
 	payload := githubSessionPayload{
 		Login:     login,
@@ -49,7 +52,11 @@ func signGitHubSession(secret, login, name, avatarURL string) (string, error) {
 	return encoded + "." + sig, nil
 }
 
-// verifyGitHubSession validates the token and returns the payload if valid and unexpired.
+// verifyGitHubSession validates a legacy token and returns the payload if
+// valid and unexpired.
+//
+// Deprecated: kept for one release so sessions issued before the JWT
+// migration keep working (dual verification in Server.verifySession).
 func verifyGitHubSession(secret, token string) (*githubSessionPayload, bool) {
 	parts := strings.SplitN(token, ".", 2)
 	if len(parts) != 2 {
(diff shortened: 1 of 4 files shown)


AnaBerg:
Phase 3.5 requires new sessions to be issued as standard JWTs and only legacy verification to remain during the transition. `signSession` still issues the old hand-rolled format whenever `LoadMasterKey` fails, using `webSessionSecret` (and therefore potentially the raw hub token) again. That turns a master-key/configuration failure into silent security downgrade and extends legacy issuance indefinitely. Please fail session issuance when the derived JWT key is unavailable; keep the old format only in `verifySession` for the one-release compatibility window.

PR #455, opened 2026-07-08
feat(hub): Prometheus /metrics endpoint and optional OTel traces

AnaBerg:
## Motivation

OTel is already in go.mod but unused and no metrics are exported; operators fly blind. Minimal Prometheus counters/histograms/gauges plus opt-in OTLP traces around provider operations.

- `go.mod` carried `go.opentelemetry.io/otel`, the SDK and the OTLP trace/metric exporters — all marked `// indirect` (pulled in by the Daytona SDK) and referenced nowhere in our code.
- No `/metrics` endpoint existed among the ~90 routes in `pkg/hub/server.go` (`registerRoutes`), and nothing exported request rates, latencies, claw counts, WebSocket traffic, webhook failures or SQLite pool stats.
- The slow operations that matter — provider `Create`/`Exec`/`Destroy` in the provisioning functions, the `terminate*VM` helpers and the workflow-run dispatch in `pkg/hub/pipeline_runner.go` — were completely untimed and untraced.

This implements re-arch spec item 1.5 (phase 1), and only that item.

## Dependencies

Depends on #453 — review and merge that first. This PR is STACKED on branch `feature/structured-logging`.

## What changed

**Prometheus `/metrics` endpoint** (`pkg/hub/metrics.go`, `pkg/hub/middleware.go`):

- Each `Server` owns its own Prometheus registry (no package globals, so the many hand-built `&Server{}` test fixtures keep working; all record methods are nil-receiver safe).
- `withMetrics` middleware wraps the ServeMux directly so it observes the matched `r.Pattern` — metrics are labeled by route pattern (e.g. `/api/claws/{id}`), never the raw path, keeping cardinality bounded. Its response recorder forwards `Hijack`/`Flush` so WebSocket upgrades keep working.
- Series exported:
  - `elasticclaw_http_requests_total{route,method,status}` and `elasticclaw_http_request_duration_seconds{route,method}`
  - `elasticclaw_claws{status}` — gauges computed from the `claws` table at scrape time (cheap `GROUP BY`), with a zero baseline for the known statuses so the series exist from first scrape
  - `elasticclaw_ws_messages_total{direction,peer}` — counted at the claw/user WebSocket read loops and the main outbound paths (broadcast, forward-to-claw, queued-message delivery, hub-injected messages)
  - `elasticclaw_webhook_errors_total{integration}` — 4xx/5xx responses on `/api/integrations/*/webhook` and `/api/workspaces/{ws}/webhooks/*` routes, derived centrally in the middleware (note: this counts rejected deliveries; handlers that swallow processing errors into a 200 are not counted)
  - `go_sql_*` pool stats via the stock `DBStatsCollector`, plus Go runtime and process collectors

**Opt-in OTel traces** (`pkg/hub/telemetry/`, `cmd/hub.go`):

- `telemetry.Setup` installs an OTLP/HTTP tracer provider only when `ELASTICCLAW_OTLP_ENDPOINT` is set (full URL or bare `host:port`, the latter sent insecure). Unset ⇒ nothing is installed, the OTel global stays no-op, `StartProviderSpan` short-circuits on an atomic bool, and the `otelhttp` middleware is not even added to the handler chain — zero cost.
- When enabled: one span per HTTP request (`otelhttp`), and manual spans around provider operations — `provider.create` in the five provisioning functions, `provider.destroy` in the five `terminate*VM` helpers, `provider.exec` around the workflow-run dispatch — each tagged `elasticclaw.provider` and recording errors.

**Dev tooling** (`docker/compose.dev.yml`, `docker/otel-collector.yaml`): commented-out `otel-collector` service plus a minimal config that prints received spans to stdout, for verifying a claw-creation trace locally.

**Dependencies** (`go.mod`): `prometheus/client_golang` and `otelhttp` added; the already-present OTel SDK modules become direct and are bumped to the versions `otelhttp` requires.

## Testing

- `go build ./...` — pass.
- `go test ./...` (what `make test` runs, without `-v`) — all packages pass, including new tests:
  - `pkg/hub/metrics_test.go`: route-pattern labeling (raw paths must not leak), latency histogram, webhook error counting, nil-safety of all record methods, `webhookIntegration` parsing.
  - `pkg/hub/telemetry/telemetry_test.go`: unset endpoint ⇒ no-op shutdown, `Enabled()` false, `StartProviderSpan` returns the identical context.
- Smoke-tested `/metrics` against a real `NewServer` + `registerRoutes` mux (temporary test, not committed): output contains `elasticclaw_http_requests_total`, `elasticclaw_http_request_duration_seconds`, `elasticclaw_claws{status=...}` and `go_sql_*` series.
- Not verified end-to-end: an actual claw-creation trace in a running collector (requires a provider account); the collector compose service was added as specced but only validated syntactically.


AnaBerg:
Manual `/metrics` validation found that the claw-status gauge is incomplete. `knownClawStatuses` only seeds:

```text
provisioning, starting, connected, offline, error, deleted
```

The generated/API status contract also includes `pending` and `idle`. On an empty hub, `elasticclaw_claws{status="pending"}` and `elasticclaw_claws{status="idle"}` are therefore absent instead of exported as zero, so dashboards cannot distinguish zero from a missing series.

Please seed every contract status and add a test that compares the metric baseline against the complete status enum.

AnaBerg:
The WebSocket metric is only incremented at a subset of write sites, so `elasticclaw_ws_messages_total{direction="out"}` materially undercounts traffic. Registration/status frames and writes in checkpoints, files, volumes, PR watcher, wake paths, and several direct server writes bypass the counter (there are many more production `wsjson.Write` calls than `wsMessage` calls). The baseline status gauge also omits `pending` and `idle`, so zero-valued series for documented statuses disappear. Please centralize WS writes behind an instrumented helper and include the complete status enum.

PR #457, opened 2026-07-08
refactor(hub): extract httpserver package (router + middlewares)

AnaBerg:
## Motivation

pkg/hub is a god package (103 files, ~54k LOC, server.go at 6,535 lines). First extraction: the HTTP layer — registerRoutes, middlewares and response helpers — behind interfaces, so handlers can migrate gradually.

Concretely, before this PR the HTTP plumbing was spread across `server.go` and `middleware.go` inside the god package:

- `Server.registerRoutes` (~130 lines) mounted ~70 routes with per-route auth wrapping, plus inline `/healthz`, e2e-bridge and `/api/debug/claws` closures.
- The middleware chain (`withRecovery`, `withRequestID`, `withMetrics`, `corsMiddleware`) and the JSON error envelope (`writeErr`/`apiError`, `jsonOK`, `jsonError`) lived next to 6k+ lines of handler, provisioning and WebSocket code, so nothing could depend on them without importing all of `pkg/hub`.
- `writeErr` alone has ~123 call sites and `jsonOK`/`jsonError` ~129 across 10+ files — the envelope is de facto shared infrastructure and needs a home that future subpackages (`analytics/`, `settings/`, …) can import without cycles.

This is item 2.2 step 1 of the re-architecture plan (docs/re-arch/phase-2-hub-reorganization.md on the `docs/re-arch` branch): a mechanical move, `git mv` for file moves, one subpackage per PR, move commit separate from the import-fix commits, zero behavior change.

## Dependencies

This PR is STACKED on branch `feature/metrics-and-traces`. Depends on #455 — review and merge that first.

## What changed

New package `pkg/hub/httpserver` (router, middleware chain, error envelope, HTTP helpers). Three commits:

1. **`git mv` only** — `pkg/hub/middleware.go` and `pkg/hub/middleware_test.go` move to `pkg/hub/httpserver/` with no content change, preserving history.
2. **Package rename + wiring** —
   - `WithRecovery`, `WithRequestID`, `WithMetrics`, `CORS`, `statusRecorder` and the envelope helpers `WriteErr`/`JSONOK`/`JSONError` become `package httpserver` exports; wire format and middleware order are unchanged.
   - `WithRequestID` now takes the base `*slog.Logger` and `WithMetrics` takes a new `httpserver.MetricsRecorder` interface, so httpserver has no dependency on hub. `serverMetrics.observeRequest`/`webhookError` are exported to satisfy it; `webhookIntegration` (route-pattern knowledge) moves to httpserver with its test.
   - hub keeps `writeErr`/`jsonOK`/`jsonError` as one-line aliases so the ~250 existing call sites do not churn; they disappear as handlers migrate to their own subpackages.
   - `TestLogfComponentExtraction` returns to `pkg/hub/logging_test.go` (it tests `logMsg`, which stays in hub).
3. **Router extraction** — the route table moves verbatim (patterns, registration order, per-route auth policy, comments) into `httpserver.RegisterRoutes(mux, auth, handlers)`. Handlers stay as unexported methods on `hub.Server`, referenced through the `httpserver.Handlers` struct; the auth middlewares (`withAuth`, `withWebAuth`, `withWebAdminAuth`, `withAdminForMethods`) stay in hub — they are entangled with Server state (token resolution, sessions, access config) — and are consumed through the `httpserver.Auth` interface via a 4-method `serverAuth` adapter. The inline `/api/debug/claws` closure becomes `Server.handleDebugClaws` so it can be wired like the rest.

Not touched (per spec, later steps): auth internals, handlers, `serveWebUI`, graceful shutdown, `Run` composition stays in hub.

## Testing

- `go build ./...` — pass.
- `go test ./...` (the `make test` suite without `-v`) — all packages pass, including `pkg/hub` (17s) and `pkg/hub/factorytest`.
- `go test -tags integration -timeout 120s ./pkg/hub/... -run TestFactory` (the `make test-factory` gate) — pass.
- `go vet ./pkg/hub/...` and `gofmt` — clean.

No test in the existing suite was changed in behavior; moved tests were only re-pointed at the exported names (`WithRecovery`, `WriteErr`, `httpserver.WithMetrics(s.metrics, …)`).


AnaBerg:
The implemented dependency direction is the reverse of the Phase 2.1 rule. The spec says `httpserver` may know service interfaces while services must not import `httpserver`; this stack has analytics/checkpoints/claws/integrations/settings/workflows importing `pkg/hub/httpserver` for response helpers, and the depguard config explicitly tolerates it. That temporary exception still exists after the Phase 2.5/#474 migration. Please move shared envelopes/helpers to a lower neutral package or keep response mapping in the HTTP adapter so service packages do not depend upward on transport code.

PR #461, opened 2026-07-08
refactor(hub): extract integrations package, shared webhook signature helper

AnaBerg:
## Motivation

Continuing the pkg/hub split (phase 2, item 2.2 step 4): linear/github/jira/shortcut webhooks and pollers move to `pkg/hub/integrations`. While moving, signature validation is extracted into one shared helper and every integration is audited to actually validate signatures (open review finding).

Concretely, the moved surface is ~9.4k LOC across `linear.go`, `github_webhook.go`, `github_issues.go`, `jira.go`, `shortcut.go`, `external_webhook.go` and `integration_poller.go`, which together accounted for ~150 `*Server` methods inside the god package. Before this PR each integration hand-rolled its own HMAC block (four copies in `validateLinearSignature` alone, plus `verifyGitHubHMAC`, `validateShortcutSignature`, `validateExternalSignatureForFactory`), with subtly different empty-signature/prefix handling.

**Signature audit result** (now documented in `pkg/hub/integrations/signature.go`; behavior unchanged):

| Integration | Mechanism | When no secret is configured |
|---|---|---|
| Linear | HMAC-SHA256, `Linear-Signature` | accepts (open webhook) |
| GitHub PR | HMAC-SHA256, `X-Hub-Signature-256` | **rejects** (strict) |
| GitHub Issues | HMAC-SHA256, `X-Hub-Signature-256` | global endpoint rejects; workspace-scoped accepts |
| Shortcut | HMAC-SHA256, `Payload-Signature` | accepts |
| Jira | shared secret (constant-time compare) — Jira cannot sign payloads | accepts |
| External | HMAC-SHA256, `X-Webhook-Signature`, per factory | factory without a secret only matches unsigned deliveries |

All six entry points do validate; the pre-existing "accept when unconfigured" defaults are preserved as-is (tightening them would be a behavior change, out of scope for a mechanical move — flagged for a follow-up if desired).

## Dependencies

Depends on #459 — review and merge that first. This PR is STACKED on branch `feature/hub-settings-package`.

## What changed

- **Move commit (git mv only):** the seven integration files move to `pkg/hub/integrations/`, preserving history. `label_match.go` (+test) and `factory_pr_policy.go` (pure trigger helpers used almost exclusively by the moved code) move along in the compile commit.
- **Compile commit:** following the analytics/settings precedent:
  - `integrations.Service` + `integrations.Deps`: all hub-owned state (DB, config mutex, claw registry, webhook dedup map, cron scheduler, pipeline/claw-lifecycle/provisioning helpers) is injected via closures; forwarding methods keep the moved call sites byte-identical wherever possible. No import of pkg/hub, so no cycle.
  - Shared cross-boundary types move with the package and are exported: `TriggerActor`, `AgentFailureFeedback/Message/Kind`, `WorkspaceIssueTracker`, `WorkflowCreateOptions`, `PRRef`/`ExtractPRs`, the factory-trigger claim sentinel.
  - `pipelineContext` moves to `pipeline.Context` (`pkg/hub/pipeline`) so both sides share it without a cycle; pkg/hub keeps a type alias.
  - `logger.Msgf` hosts the shared printf→slog bridge so hub and integrations keep identical log formatting.
  - `pkg/hub/integrations_bridge.go` keeps every existing pkg/hub call site (including all tests, which stay in pkg/hub per the settings precedent) unchanged via type aliases, package-level aliases, and thin `*Server` wrappers.
- **Signature commit:** one shared `verifyHMACSHA256` helper (constant-time, handles the `sha256=` prefix, never matches empty secrets/signatures) replaces the six hand-rolled HMAC blocks; `constantTimeStringEqual` (Jira) moves next to it; the audit table above is documented in the file header.

No behavior changes intended; no CI workflow changes.

## Testing

```
go build ./...                                              # OK
go vet ./...                                                # OK
go test ./...                                               # OK (full suite, all packages pass)
go test -count=1 -tags integration ./pkg/hub/ -run TestFactory   # OK (make test-factory equivalent)
go test -tags integration -timeout 300s ./pkg/hub/ -run TestParity  # OK (make test-parity equivalent)
go test -count=1 -race ./pkg/hub/integrations/              # OK
```

All commands run locally on this branch; no failures.


AnaBerg:
The signature audit documents that several public webhook routes still accept unsigned requests when no secret is configured: Linear, Shortcut, Jira, and workspace-scoped GitHub Issues. The implementation confirms this with `return true` / `return !hasSecrets` fallbacks. That does not satisfy Phase 2.2's requirement to ensure every integration validates webhook authenticity, and it leaves claw-creation endpoints open by default. Please fail closed when no verification secret is configured (with a clear configuration error), and remove the Jira `?secret=` fallback so webhook credentials do not leak through URLs.

PR #469, opened 2026-07-08
refactor(hub): extract claws package (WS, streaming, files, terminal)

AnaBerg:
## Motivation

Continuing the pkg/hub split (phase-2 re-architecture, item 2.2 step 7): the claw lifecycle core — WS handling, streaming, files, terminal — moves to `pkg/hub/claws`. Extracted last per the spec because it is the highest-risk piece: `handleClawWS` alone was ~600 lines inside the 6,605-line `server.go`, owning bridge registration, the status channel, heartbeats (gateway promotion, unhealthy tracking, context-usage warnings), streaming-chunk buffering with partial-message upserts, message finalization with [DONE]/[TERMINATE] signal handling, file/volume ack resolution, and the reverse HTTP proxy — all interleaved with the `clawConn` per-connection mutex and the server registry mutex. After this step, `server.go` drops from 6,605 to ~5,570 lines, leaving composition/lifecycle for step 8 (`hub.go`).

## Dependencies

This PR is STACKED on branch `feature/hub-workflows-package`. Depends on #468 — review and merge that first.

## What changed

- **Commit 1 (git mv only):** `files.go` (claw file upload/view over the claw WS channel) moves to `pkg/hub/claws/files.go`, preserving history. It is the only whole-file move; the rest of the claws surface lived inside `server.go` and cannot be `git mv`'d.
- **Commit 2 (compile + bridge):**
  - `clawConn` moves to `claws.Conn` with exported fields; pkg/hub keeps it via `type clawConn = claws.Conn`, and every remaining hub-side field access (user-WS fanout, status watchdog, wake glue, hand-built test servers) is a mechanical case rename on the same struct, same mutex. The `checkpoints.Conn`/`workflows.Conn` accessor methods move from the hub bridges to `claws/conn.go` unchanged, so `*claws.Conn` keeps satisfying both dependency interfaces.
  - `handleClawWS`, `flushStreamingSegment`, `handleTerminal`, `injectHubMessage`, `sendNextQueuedMessage` and the files handlers become `claws.Service` methods with the same bodies. A rename-normalized diff of the old vs. new `handleClawWS` shows exactly two differences: the exported field names in the `Conn` literal, and the pipeline-trigger block (below).
  - New `claws.Service` + `Deps` follow the established analytics/settings/integrations/checkpoints/workflows pattern: hooks for the hub DB, registry mutex + map, waiter maps under the shared `fileAckMu`, claw-token auth, broadcast, checkpoint/volume bridges, initial-plan handling, done/terminate signals, and SSH identity. The registry itself stays on the hub `Server` for now (zero behavior change; its dedicated RWMutex is spec item 2.3).
  - The inline pipeline `[DONE]`/message trigger block of `handleClawWS` becomes `Server.evaluatePipelineMessageTriggers` (same statements, relocated to pkg/hub) reached through a Deps hook, because it builds workflows-package pipeline contexts and the dependency rules say claws must not import workflows.
  - `pkg/hub/claws_bridge.go` keeps every existing call site (routes, sibling bridges, tests) compiling unchanged via the type alias, var aliases (`detectToolLoop`, `allowWakeBeforeBootstrap`, activity-payload helpers) and thin delegating methods.
- Tests stay in pkg/hub (they exercise hub `Server`/`clawConn` internals), matching all previous extractions; test changes are limited to the exported-field renames.

## Testing

- `go build ./...` — pass
- `go vet ./...` — pass
- `go test ./...` (make test without -v) — all packages pass (`pkg/hub` 15.3s)
- `make test-parity` equivalent: `go test -tags integration -timeout 300s ./pkg/hub/ -run TestParity` — pass
- `go test -tags integration -timeout 120s ./pkg/hub/ -run TestFactory` — pass
- `go test -race -timeout 300s ./pkg/hub/ ./pkg/hub/claws/` — pass (36s)
- Fidelity check: extracted the pre-move `handleClawWS` from git, applied the mechanical renames with a script, and diffed against the moved body — only the two expected differences.


AnaBerg:
The documented `claws/` ownership boundary is only partially implemented. Phase 2.1 assigns claw lifecycle, WS connections, files, and streaming to this package, but this PR moves only WS/streaming/files/terminal. The collection/detail lifecycle handlers and all provisioning/termination logic remain on `hub.Server` in `server.go` (for example `handleClaws`, `handleClawDetail`, `provision*`, and `terminate*`), and the following composition-root PR merely splits them into root-package files.

As a result, `pkg/hub/claws` still does not own claw lifecycle and the root hub package remains a domain implementation rather than composition. Please move the lifecycle service/handlers behind the claws package boundary and leave only wiring/adapters in the root package.

PR #470, opened 2026-07-08
refactor(hub): reduce server.go to a composition root (hub.go)

AnaBerg:
## Motivation

Final step of the split: the residual server.go becomes hub.go, holding only composition (store -> services -> httpserver) and lifecycle Run(ctx). Add depguard rules to golangci config to enforce the dependency direction between subpackages.

After steps 1–7 extracted `httpserver/`, `analytics/`, `settings/`, `integrations/`, `checkpoints/`, `workflows/`, and `claws/`, `server.go` was still 5,576 LOC. Only ~280 of those lines are actual composition (the `Server` struct, `NewServer` wiring db → services → background loops, `Run`/`registerRoutes`, and the `serverAuth` adapter for the `httpserver` route table); the rest is auth middleware, claw/message HTTP handlers, the user WebSocket fan-out, and ~2.7k LOC of per-provider provisioning/bootstrap code. This PR renames the file to `hub.go` and moves everything that is not composition into thematic files, all within `package hub` — a pure mechanical file split, no identifier changes, no behavior changes.

## Dependencies

This PR is STACKED on branch `feature/hub-claws-package`. Depends on #469 — review and merge that first.

## What changed

- `git mv pkg/hub/server.go pkg/hub/hub.go` (history-preserving rename, own commit).
- `hub.go` reduced to the composition root and lifecycle: **281 LOC** (target <500). The rest moved verbatim into new files in the same package:
  - `auth.go` — auth middlewares, token/tenant resolution, login/session/GitHub-token handlers (474 LOC)
  - `webui.go` — embedded web UI serving, branding/hub-config endpoints (188)
  - `claw_handlers.go` — claw collection/detail HTTP handlers (562)
  - `message_handlers.go` — conversation message/timeline/activity handlers and queries (520)
  - `userws.go` — user WebSocket connections, broadcast fan-out, JSON response helpers (282)
  - `provision.go` — provider-agnostic provisioning dispatch, bootstrap status, shared helpers (378)
  - `provision_daytona.go` / `provision_replicated.go` / `provision_exedev.go` / `provision_docker.go` — per-provider provisioning, bootstrap, and teardown (961 / 560 / 206 / 394)
  - `background.go` — provider status polling and the status watchdog (146)
  - `wake.go` — wake messages and the initial-plan protocol (188)
  - `clawenv.go` — env/shell-script builders injected into claw sandboxes (450)
  - `ssh.go` — SSH exec/file-write helpers for remote VMs (184)
- `git mv pkg/hub/server_test.go pkg/hub/hub_test.go`, then split the message/timeline tests into `message_handlers_test.go` (660) and the initial-plan tests into `wake_test.go` (111), bringing `hub_test.go` to 1,201 LOC (was 1,954; the ceiling is 1,500).
- `.golangci.yml`: enabled `depguard` with the 2.1 dependency rules:
  - `httpserver` may not import any service package (`claws`, `integrations`, `workflows`, `checkpoints`, `analytics`, `settings`) nor a future `store` — the route table takes handlers via interfaces.
  - `claws` (the core) may not import `integrations` or `workflows`; those may call `claws`, never the reverse.
  - `store` (once it exists, phase 2.4) may only know `pkg/types`.

### Honest residuals vs the quantitative target

- `hub.go` is 281 LOC (<500 ✓) and every file touched by this PR is under the 1,500 LOC ceiling ✓.
- Files still over 1,500 LOC elsewhere in the repo, out of scope for step 8 (they belong to the packages extracted in earlier steps): `pkg/hub/workflows/pipeline_runner.go` (2,069), `pkg/hub/integrations/linear.go` (1,947), `pkg/hub/integrations/integration_poller.go` (1,594), `pkg/hub/analytics/task_run_analytics.go` (1,565), `pkg/hub/settings/settings.go` (1,562), and `cmd/claw-bridge/main.go` (3,208, not part of `pkg/hub`).
- The spec's "services do not know httpserver" rule is only partially enforceable today: service packages import `httpserver` solely for the shared response helpers (`JSONOK`/`JSONError`/`WriteErr`). This is documented in the config comment and becomes fully enforceable when the error-envelope mapping is centralized (phase 2.5). The rules added here enforce the directions that must never regress.
- `NewServer` still spawns background goroutines and `Run` still uses `http.ListenAndServe`; graceful shutdown/lifecycle changes are Phase 0/2.3 items, not touched here.

## Testing

- `go build ./...` — pass.
- `go vet ./pkg/hub/...` — pass.
- `go test ./...` (the `make test` command minus `-v`) — **all packages pass**, including `pkg/hub` (16.7s).
- `go test -tags integration -timeout 60s ./pkg/hub/... -run TestFactory` (`make test-factory`) — pass.
- `golangci-lint run ./pkg/hub/...` (v2.1.6) — **no depguard or forbidigo violations**. The pre-existing default-linter findings (errcheck/staticcheck/unused; 114 across `pkg/hub`) are unchanged by this PR: every finding located in the new files points at code moved verbatim from `server.go`.
- `gofmt -l pkg/hub` — clean.


pkg/hub/hub.go
```diff
@@ -0,0 +1,281 @@
...
+	db        *sql.DB
+	addr      string
+	logger    *slog.Logger
+	hubCfg    *types.HubConfig
+	identity  *HubIdentity
+	mux       *http.ServeMux
+	artifacts artifact.Store
+	metrics   *serverMetrics
+
+	mu    sync.RWMutex
```
AnaBerg, line 32: docs/re-arch spec-alignment: Phase 2.3 says there should be one mutex per subsystem and that no global lock remains in `hub.go`. This composition root still owns a single `s.mu` alongside `claws`, `users`, and hub config state, so the split preserves the concurrency boundary the phase is meant to remove. Please move these registries and their locks into the subsystem services, or otherwise introduce narrow service-owned mutexes before treating this as the residual composition root.

pkg/hub/claw_handlers.go
```diff
@@ -0,0 +1,562 @@
...
+	// Use a stable short ID as the provider-side name so renaming the claw
+	// doesn't require a provider API call.
+	providerNamePrefix := strings.TrimSpace(os.Getenv("ELASTICCLAW_PROVIDER_NAME_PREFIX"))
+	if providerNamePrefix == "" {
+		providerNamePrefix = "ec-"
+	}
+	req.ProviderName = providerNamePrefix + clawID[:8]
+	go func() {
+		logfCtx(r.Context(), "Provisioning claw %s (%s) via %s (provider name: %s)...", req.Name, clawID, req.Provider, req.ProviderName)
+		ctx := context.Background()
```
AnaBerg, line 301: docs/re-arch spec-alignment: Phase 2.3 requires work that outlives an HTTP request to derive from the server's root context with an explicit timeout, never bare `context.Background()`. This provisioning goroutine is exactly that out-of-request case; `Background()` detaches it from hub shutdown and gives provider/bootstrap work no deadline. Please derive a provisioning context from the hub/root context with a bounded timeout and pass that through the provider calls.

pkg/hub/provision_daytona.go
```diff
@@ -0,0 +1,961 @@
...
+		// Each step inside bootstrapDaytona retries 3x internally.
+		// Outer retries here handle the rare case of total step failure.
+		const maxBootstrapAttempts = 3
+		var lastErr error
+		for attempt := 1; attempt <= maxBootstrapAttempts; attempt++ {
+			if attempt > 1 {
+				logfCtx(ctx, "[daytona] full bootstrap retry for claw %s in 15s...", clawName)
+				time.Sleep(15 * time.Second)
+			}
+			lastErr = s.bootstrapDaytona(context.Background(), clawID, clawName, instance.ID, p, env)
```
AnaBerg, line 61: docs/re-arch spec-alignment: even if the caller passes a lifecycle-aware context into `provisionDaytona`, this bootstrap goroutine discards it and switches back to `context.Background()`. That defeats the Phase 2.3 context sweep for long-running provisioning work. Please derive per-attempt contexts from the incoming/root context, ideally with bootstrap step timeouts, and reuse them for both `bootstrapDaytona` and the bridge-running check.

pkg/hub/message_handlers.go
```diff
@@ -0,0 +1,520 @@
...
+				writeErr(w, http.StatusForbidden, "forbidden", "forbidden")
+				return
+			}
+		}
+
+		msg := types.HubMessage{
+			ID: uuid.New().String(), ClawID: clawID, TenantID: tenantID,
+			Role: "user", Content: body.Content, CreatedAt: now(),
+		}
+		if _, err := s.db.Exec(
```
AnaBerg, line 83: docs/re-arch spec-alignment: Phase 2.4 says raw `db.Exec`/`Query` calls should move behind `store/` repositories, with multi-step operations handled transactionally. This handler still inserts a message directly and then records task-run analytics separately, so a partial failure can leave message/analytics state inconsistent and keeps SQL in the handler layer. Please route this through a messages/task-run repository method, or a `store.WithTx` operation that covers the insert plus related analytics update.

AnaBerg:
The current "final step" still misses two explicit Phase 2 end-state constraints:

- The quantitative target says no `pkg/hub` file should exceed 1,500 LOC. This head still has `workflows/pipeline_runner.go` (2,069), `integrations/linear.go` (1,947), `integrations/integration_poller.go` (1,594), `analytics/task_run_analytics.go` (1,565), and `settings/settings.go` (1,562).
- The dependency rule says services must not know `httpserver`, but analytics/checkpoints/claws/settings/workflows still import `pkg/hub/httpserver` for response helpers, and the depguard configuration explicitly tolerates that direction.

The PR description acknowledges both as residuals, but that means this is not yet the documented final architecture. Please finish the remaining splits and invert/move the response helpers so the rule can actually be enforced, or stop presenting this PR as completion of the final composition-root step.

AnaBerg:
This PR is the final composition-root step of the extraction roadmap, but the quantitative end-state from Phase 2.2 is not reached. At this head, five production files still exceed 1,500 lines: `workflows/pipeline_runner.go` (2,069), `integrations/linear.go` (1,947), `integrations/integration_poller.go` (1,594), `analytics/task_run_analytics.go` (1,565), and `settings/settings.go` (1,562). Please split these remaining oversized units before treating the reorganization as complete.

pkg/hub/hub.go
AnaBerg: Agreed that the global `s.mu` violates the Phase 2.3 target — but per the spec's one-item-per-PR rule, this PR is the Phase 2.2 step 8 mechanical split (no behavior change), and the mutex-per-subsystem work is Phase 2.3, implemented in the descendant stacked PR #472: the global lock is removed there in favor of `claws.Registry` (own RWMutex), `userRegistry` and `cfgMu`, with the lock-acquisition order documented in `hub.go`. Applying it here would duplicate #472 and break this PR's mechanical-move property, so it stays as-is on this link of the stack.

pkg/hub/claw_handlers.go
AnaBerg: Correct observation — this goroutine still uses bare `context.Background()` on this branch. The context sweep is Phase 2.3 item 3, and it is implemented in the descendant stacked PR #472, where this provisioning goroutine derives its context from the hub root context (`ctx := s.base()`). This PR is the Phase 2.2 step 8 mechanical split, so the code was moved verbatim; fixing it here would conflict with #472.

pkg/hub/provision_daytona.go
AnaBerg: Same situation as the sibling comment: the bootstrap-retry goroutine's `context.Background()` calls are Phase 2.3 item 3 work, already reworked in descendant PR #472 (each attempt derives from the hub root context via `s.base()`, with an explanatory comment). Kept verbatim here to preserve this PR's mechanical-move property.

pkg/hub/message_handlers.go
AnaBerg: Correct — the raw `db.Exec` insert is still here, but the `store/` layer is Phase 2.4, which is exactly the next PR in the stack: #471 routes this insert through `s.st().Messages().Insert(r.Context(), msg)`. One genuine residual remains on #471's side and is noted for its review: the message insert and the task-run analytics recording are still separate calls rather than one `store.WithTx` transaction. That follow-up belongs on #471, not this mechanical-split PR.

PR #472, opened 2026-07-09
refactor(hub): per-subsystem mutexes, WS worker pool, context propagation

AnaBerg:
## Motivation

A single `Server` mutex guards claws, users, config and caches; each WS message spawns an unbounded goroutine; 84 `context.Background()` calls sever cancellation. Per-subsystem locks, a semaphore-bounded worker pool (default 64), `r.Context()`/root-ctx derivation, and periodic partial persistence for streaming.

- `Server.mu` (`pkg/hub/hub.go`) was one RWMutex for three unrelated subsystems: the claw-connection map, the user-session map, and the live `hubCfg` pointer. Every subpackage extracted in the earlier phase-2 steps received `Mu *sync.RWMutex` pointing back at that same lock (`claws/deps.go`, `workflows/deps.go`, `integrations/deps.go`, `checkpoints/deps.go`, `settings/service.go`), so a config read in a webhook handler contended with WS registration and user fanout.
- The claw WS read loop (`pkg/hub/claws/ws.go`, formerly `server.go:2546`) spawned `go func(payload, conn)` per `http_proxy_req` and per-message goroutines for `[DONE]`/`[TERMINATE]` handling, PR scanning and hub message injection — no bound, so a looping bridge could exhaust the hub.
- 77 non-test `context.Background()` calls remained in `pkg/hub` (down from the 84 in the review after the store-layer extraction), severing cancellation in handlers and detaching all background work from any root context.
- The streaming chunk path upserted the entire accumulated buffer to SQLite on every chunk — durable, but O(n²) write amplification over a long response — and the only mid-stream loss protection was the disconnect flush.

## Dependencies

This PR is STACKED on branch `feature/hub-store-layer`. Depends on #471 — review and merge that first.

## What changed

1. **One mutex per subsystem** (commit `refactor(hub): split global Server mutex…`)
   - `claws.Registry`: claw_id → `Conn` map with its own RWMutex, owned by `pkg/hub/claws`; compound read-modify-write operations (re-registration queue migration, disconnect flush, kill-and-delete) go through `Do`/`DoRead` so they keep their pre-split atomicity.
   - `userRegistry` (`pkg/hub/userws.go`): user-session map with its own RWMutex.
   - `Server.cfgMu`: guards only the `hubCfg` pointer; the subpackage deps field is renamed `Mu` → `CfgMu` so its scope is explicit.
   - `Server.markerMu`: the wake system-marker check+insert kept its atomicity on a dedicated mutex.
   - No global lock remains; lock-ordering rules are documented on the `Server` struct (subsystem locks are leaves; registry lock → `Conn.Mu` only inside registry callbacks).

2. **Semaphore-bounded WS worker pool** (`pkg/hub/claws/pool.go`)
   - `claws.WorkerPool` wraps `semaphore.Weighted` (x/sync was already a dependency); default 64 concurrent handlers per hub, configurable via `ws_worker_limit` in `hub.yaml` (`types.HubConfig.WSWorkerLimit`).
   - Every per-message goroutine in `HandleClawWS` (http proxy, `[DONE]`/`[TERMINATE]`, PR scan, message injection, bootstrap checkpoint) now goes through the pool. On overflow the connection is closed with the dedicated close code **4429 (`StatusOverloaded`)**, so a malicious or looping client applies backpressure to itself.

3. **`context.Background()` sweep** — HTTP-handler paths now use `r.Context()`; work that must outlive a request (provisioning continuations, disconnect cleanup, pollers, checkpoints, model-auth login job, metric scrapes, singleflight refresh) derives from the new hub root context `Server.baseCtx` through a nil-safe `BaseCtx` hook injected into the claws/checkpoints/workflows/integrations/settings services. Existing `WithTimeout` sites keep their timeouts, now parented on the root. The **`contextcheck` linter is enabled** in `.golangci.yml`; intentional detach sinks are excluded by name with justification, and the handful of inline cases carry `//nolint:contextcheck` comments.

4. **Streaming partial commits** — the per-chunk full-buffer upsert became a thresholded partial commit (first chunk, then every 8 KiB or 2 s) under the same message ID; the finalizing message or the disconnect flush (which appends `[interrupted]`) overwrites it. A mid-stream disconnect preserves everything received; a hub crash loses at most one flush window.

5. **Acceptance artifacts**
   - `hack/ws-load-test`: connects 100 simulated claws, keeps them streaming, samples `go_goroutines` from `/metrics`, exits non-zero on unbounded growth.
   - `TestMidStreamDisconnectPreservesReceivedContent` (`pkg/hub/streaming_disconnect_test.go`): kills the claw WS mid-stream and asserts the received content is persisted and marked interrupted.
   - `TestWorkerPoolBoundsConcurrency` / `TestWorkerPoolZeroValueUsesDefaultLimit` (`pkg/hub/claws/pool_test.go`).

Out of scope (unchanged): graceful shutdown itself (the root context is still `context.Background()` today — it becomes cancelable when the phase-0 shutdown work lands and everything derived here stops with it), and threading request contexts through the remaining bridge hooks (excluded sinks are documented in `.golangci.yml`).

## Testing

- `go build ./...` — pass.
- `go test ./...` (the `make test` suite, non-verbose) — pass, no failures.
- `go test -race ./pkg/hub/...` — pass (mandatory per the spec's split-mutex risk mitigation).
- `go test -tags integration -timeout 120s ./pkg/hub/... -run TestFactory` (make test-factory) — pass.
- `go test -tags integration -timeout 300s ./pkg/hub/ -run TestParity` (make test-parity) — pass.
- `golangci-lint run` — zero `contextcheck` findings; remaining findings (errcheck/staticcheck/unused) pre-date this PR.
- Load test executed against a locally built hub: `go run ./hack/ws-load-test -hub http://127.0.0.1:18095 -claw-token … -claws 100 -duration 45s` → `claws connected=100 failed=0 baseline=215 final=215 / PASS: goroutine count stable`.


AnaBerg (commented):
Posting the seven pending inline comments from the prior CodeRabbit/docs-re-arch review. No new findings added.

pkg/hub/claws/pool.go
```diff
@@ -0,0 +1,58 @@
...
+// the queue overflowed — and the caller should apply backpressure (the WS
+// read loop closes the connection with StatusOverloaded).
+func (p *WorkerPool) TrySubmit(fn func()) bool {
+	p.init()
+	if !p.sem.TryAcquire(1) {
+		return false
+	}
+	go func() {
+		defer p.sem.Release(1)
+		fn()
```
AnaBerg, line 65: CodeRabbit (major): `TrySubmit` runs caller-provided work in a goroutine without panic containment. A panic from one WS handler can still bring down the hub process even though the semaphore permit is released. Wrap the goroutine body with a deferred `recover` guard around `fn()` while keeping `defer p.sem.Release(1)` so the worker pool contains handler panics instead of propagating them process-wide.

hack/ws-load-test/main.go
```diff
@@ -0,0 +1,200 @@
...
+	defer chatter.Stop()
+	turn := 0
+	for {
+		select {
+		case <-stop:
+			return nil
+		case <-ctx.Done():
+			return nil
+		case <-heartbeat.C:
+			_ = wsjson.Write(ctx, conn, wsMessage{Type: "heartbeat", Payload: map[string]interface{}{
```
AnaBerg: CodeRabbit (minor): the load-test claw treats write/read failures as success. The drain goroutine exits silently on `wsjson.Read` errors and these writes discard errors, so a dead connection can still be counted as connected. Please propagate reader/writer errors back to `runClaw` (for example via an error channel), stop the writer when the reader exits, and return a non-nil error for failed connections.

hack/ws-load-test/main.go
```diff
@@ -0,0 +1,200 @@
...
+			_ = wsjson.Write(ctx, conn, wsMessage{Type: "message", Payload: map[string]string{
+				"content": fmt.Sprintf("turn %d complete from claw %03d", turn, i),
+			}})
+			turn++
+		}
+	}
+}
+
+func scrapeGoroutines(hubURL string) (float64, error) {
+	resp, err := http.Get(strings.TrimRight(hubURL, "/") + "/metrics")
```
AnaBerg: CodeRabbit (minor): this scrape can block forever if the hub accepts the TCP connection but stalls, which would hang the load test instead of reporting a sample failure. Use an `http.Client` with a finite timeout in `scrapeGoroutines` and keep the existing status-code handling.

pkg/hub/claws/ws.go
```diff
@@ -90,64 +90,72 @@ func (s *Service) HandleClawWS(w http.ResponseWriter, r *http.Request) {
...
 			_ = wsjson.Write(ctx, conn, types.WSMessage{Type: "registered", Payload: map[string]string{"claw_id": clawID, "channel": "status"}})
 			// Simple read loop for status channel — just keepalive
 			for {
 				var msg types.WSMessage
 				if err := wsjson.Read(ctx, conn, &msg); err != nil {
-					s.mu.Lock()
-					if existing2, ok2 := s.claws()[clawID]; ok2 {
+					if existing2, ok2 := s.reg.Get(clawID); ok2 {
 						existing2.Mu.Lock()
 						existing2.StatusConn = nil
```
AnaBerg: CodeRabbit (minor): this cleanup unconditionally clears `StatusConn` for the claw ID. If the status channel reconnects before the old read loop returns, the old goroutine can erase the newer connection. Only set `StatusConn = nil` when the current value still matches this `conn` instance.

pkg/hub/checkpoints/deps.go
```diff
@@ -59,10 +59,16 @@ type ConnEntry struct {
...
+	// outlive an HTTP request derives from it — never from a bare
+	// context.Background() — so cancellation propagates once graceful
+	// shutdown wires a cancelable root. May be nil in hand-built test
+	// servers; use the service's baseCtx() accessor.
+	BaseCtx func() context.Context
 	// Mu is the hub's config/claw-registry mutex. It must be the same
 	// mutex the hub uses so reads/writes keep the exact same
 	// synchronization as before the extraction.
-	Mu *sync.RWMutex
+	CfgMu *sync.RWMutex
```
AnaBerg, line 71: CodeRabbit (minor): the field was renamed to `CfgMu`, but the surrounding comment still says `Mu` is the config/claw-registry mutex. Since claw connections now go through the claw registry bridge, update this comment and the nearby `HubCfg` comment to describe only the config mutex role.

pkg/hub/integrations/deps.go
```diff
@@ -21,10 +21,16 @@ import (
...
+	// outlive an HTTP request derives from it — never from a bare
+	// context.Background() — so cancellation propagates once graceful
+	// shutdown wires a cancelable root. May be nil in hand-built test
+	// servers; use the service's baseCtx() accessor.
+	BaseCtx func() context.Context
 	// Mu is the hub's config/claw-registry mutex. It must be the same
 	// mutex the hub uses so reads/writes keep the exact same
 	// synchronization as before the extraction.
-	Mu *sync.RWMutex
+	CfgMu *sync.RWMutex
```
AnaBerg, line 33: CodeRabbit (minor): this field is now `CfgMu`, but the comment still refers to `Mu` and describes the old config/claw-registry lock. Please update this and the nearby `HubCfg` comment so the extracted `Deps` API documents the current mutex name and scope.

.golangci.yml
```diff
@@ -61,6 +66,16 @@ linters:
...
+      # contextcheck: these sinks intentionally run on the hub root context
+      # (Server.baseCtx) because their work outlives any single request —
+      # user-WS fanout, queued message delivery, checkpoints, shared
+      # singleflight refreshes, cleanup after provisioning failures.
+      # Threading request contexts through the bridge hooks they sit behind
+      # is part of the remaining phase-2 extraction steps; new bare
+      # context.Background() call sites are still flagged.
+      - linters:
+          - contextcheck
+        text: "(sendNextQueuedMessage|clawTagsForBroadcast|tenantByToken|hasSystemMarker|clawHasMessages|CheckpointBeforeTermination|provisionStoredClaw|DrainPendingCheckpoint|snapshot\\$1|checkFactories|runModelAuthLoginJob|loadHubTemplate|handleTroubleshootStream|createClawFromFactory|cleanupClaw|HandleClawWS\\$|flushStreamingSegment|createClawForExternalEvent|buildGitHubIssuesContextWithComments|createClawForGitHubPR|ProvisionPendingClaw|setReady|checkGateway)"
```
AnaBerg: docs/re-arch spec-alignment: Phase 2.3 calls out enabling `contextcheck` specifically to prevent new bare `context.Background()` regressions in HTTP/WS call chains. This exclusion suppresses the entire `HandleClawWS` path, which is the main WS worker-pool and root-context migration surface. Please remove `HandleClawWS\$` from the broad regex, or replace it with tightly scoped `//nolint:contextcheck` annotations for the specific bridge hooks that intentionally outlive the request.

pkg/hub/claws/ws.go
```diff
@@ -167,29 +175,31 @@ func (s *Service) HandleClawWS(w http.ResponseWriter, r *http.Request) {
...
 		var partialContent string
 		var partialMsgID string
-		// Flush any partial streaming buffer as an interrupted message
-		if partialCC, ok := s.claws()[clawID]; ok && partialCC.StreamingBuf.Len() > 0 {
-			partialContent = partialCC.StreamingBuf.String() + " [interrupted]"
-			partialMsgID = partialCC.StreamingMsgID
-			if partialMsgID == "" {
-				partialMsgID = uuid.New().String()
+		s.reg.Do(func(conns map[string]*Conn) {
+			// Flush any partial streaming buffer as an interrupted message
```
AnaBerg: manual review: after the mutex split, the registry lock only protects the map, not `Conn` fields. This disconnect cleanup reads and resets `StreamingBuf`/`StreamingMsgID` without `partialCC.Mu`, while chunk/finalization paths mutate the same fields under `Conn.Mu`; `go test -race ./pkg/hub -run TestMidStreamDisconnectPreservesReceivedContent -count=1` reports a race here (reset at ws.go:195). Please take `partialCC.Mu` before inspecting/resetting the streaming state, or snapshot the connection and then lock it before touching these fields.

pkg/hub/claw_handlers.go
```diff
@@ -34,21 +33,19 @@ func (s *Server) handleClaws(w http.ResponseWriter, r *http.Request) {
...
 	var out []types.Claw
 	for _, c := range dbClaws {
 		c.GitHubIssueURL = githubIssueURL(c.GitHubIssueID)
 		c.TenantID = tenantID
-		s.mu.RLock()
-		cc, online := s.claws[c.ID]
-		s.mu.RUnlock()
+		cc, online := s.clawReg.Get(c.ID)
 		if online {
 			// Claw is currently connected — show live status
```
AnaBerg: manual review: `clawReg.Get` only protects the map lookup; it does not protect live fields on the returned `Conn`. `GatewayReady` and `ContextUsage` are written by the heartbeat path under `cc.Mu`, so reading them here without `cc.Mu.RLock()` introduces a race after the global mutex split. The same pattern appears in the claw detail/user-WS initial status snapshots and `ClawGatewayReady`; please snapshot those live fields under the per-connection lock.

pkg/hub/claws_bridge.go
```diff
@@ -35,8 +35,10 @@ type clawConn = claws.Conn
 func (s *Server) clawsSvc() *claws.Service {
 	return claws.New(claws.Deps{
 		DB:              s.db,
-		Mu:              &s.mu,
-		Claws:           func() map[string]*clawConn { return s.claws },
+		BaseCtx:         s.base,
+		Registry:        &s.clawReg,
+		Pool:            &s.wsPool,
```
AnaBerg, line 40: manual review: wiring the `WorkerPool` into the claws service does not bound the `EvaluatePipelineMessageTriggers` hook below: that hook is invoked for each finalized claw WS message and still starts pipeline work with raw `go` calls (`transitionPipelineStageWithContext` / `checkPipelineMessageTriggers`). A chatty or looping bridge can therefore bypass the semaphore even though the rest of `HandleClawWS` now goes through `submit`. Please route those async pipeline jobs through the same bounded pool, or make the hook synchronous and have the caller submit the work.

AnaBerg (commented):
manual review: found three new actionable issues in the concurrency/goroutine-bounds paths.

AnaBerg:
The context-propagation work is not connected to a cancelable hub lifecycle in this stack. `NewServer` still initializes `baseCtx := context.Background()`, and the service fallbacks also return `context.Background()`. The PR description explicitly says cancellation will only exist after the graceful-shutdown work lands, but this PR is stacked on #471 rather than on the Phase 0 shutdown branch.

That conflicts with the Phase 2 dependency rule (shutdown is a prerequisite) and with item 2.3's requirement that work which outlives a request derive from the server root context so it stops with the hub. In the current branch, propagating `baseCtx` cannot cancel any of that work.

Please integrate/rebase onto the graceful-shutdown root context (or introduce an equivalent cancelable lifecycle here) and add a signal/shutdown test proving the propagated work actually stops.

AnaBerg:
Disconnect cleanup is keyed only by `clawID`, not by the connection being closed. If a claw reconnects before the old handler's defer runs, `conns[clawID]` points to the new `Conn`; the old handler then flushes/resets the new stream and unconditionally deletes the new registry entry. This can make a healthy replacement connection disappear or persist it as interrupted. Please compare the registry entry with this handler's `cc`/`conn` before mutating or deleting it, and leave a newer connection untouched.

AnaBerg:
The context sweep does not provide runtime cancellation because `NewServer` still sets `baseCtx := context.Background()` and the entire Phase 2 stack has no `Shutdown`/cancel path. Replacing bare backgrounds with derivatives of an immortal background context only satisfies the linter syntactically; provisioning, pollers, and pool work still cannot be canceled on shutdown. Phase 2 explicitly depends on the Phase 0 graceful-shutdown root. Please integrate a caller-supplied/cancelable root context and wait for the WS worker pool before closing the database.

PR #473, opened 2026-07-09
refactor(hub): sentinel domain errors and typed WS payload decoding

AnaBerg:
## Motivation

Handlers map errors to statuses ad hoc, and `WSMessage.Payload` is `interface{}` with panicky type assertions. Sentinel errors in `pkg/types` with a single `errors.Is` -> status map in `httpserver`, and a payload decoder registry unmarshalling into concrete types (wire envelope unchanged).

- The same "not found" condition produced different statuses and bodies depending on the endpoint: `restoreClawFromCheckpoint` returned `fmt.Errorf("checkpoint not found")` and the handler blanket-mapped every restore error to **409 Conflict**; the terminal proxy wrote plaintext `http.Error(w, "not found", 404)`; workspace workflow lookups mixed plaintext `http.Error` and the JSON `jsonError` envelope for the identical "workflow not found" case (`workspace_api.go:214/253/296`).
- Handlers checked `sql.ErrNoRows` inline at each call site (`claw_handlers.go`, `message_handlers.go`, `claws/terminal.go`, ...) and hand-picked a status each time.
- Every WS read path decoded payloads by round-tripping through `interface{}`: `payload, _ := json.Marshal(msg.Payload)` followed by `json.Unmarshal` into an anonymous struct — in the claw bridge read loop (9 message types in a 400-line if/else chain in `claws/ws.go`), the user WS loop (`userws.go`), and the CLI client (`client.go`). Malformed payloads for known types were silently dropped (`if err == nil { ... }`), hiding protocol bugs.

## Dependencies

This PR is STACKED on branch `feature/hub-ws-worker-pool`. Depends on #472 — review and merge that first.

## What changed

**1. Sentinel domain errors + single status map (spec item 2.5.1)**

- New `pkg/types/errors.go`: `ErrUnauthorized`, `ErrTenantMismatch`, `ErrClawNotFound`, `ErrWorkflowNotFound`, `ErrWorkspaceNotFound`, `ErrCheckpointNotFound`, `ErrCheckpointNotReady`.
- New `pkg/hub/httpserver/errors.go`: `StatusForError(err) (status, code)` resolves the sentinel table with `errors.Is` (sees through `%w` wrapping); `WriteDomainErr(w, err)` writes the unified JSON envelope. Unrecognized errors become a generic 500 so internal detail (SQL text, paths) never leaks into responses.
- Adopted in the handlers that mapped these conditions ad hoc: checkpoint restore, the SSH terminal proxy, workspace workflow/workspace lookups, claw detail, and the message tag check. Deliberate status corrections that fall out of the single map: checkpoint restore now returns **404** for a missing checkpoint and **409** only for not-ready (previously both were 409, and a DB error was also 409 — now 500).

**2. Typed WS payload decoding (spec item 2.5.2)**

- New `types.WSEnvelope{Type string; Payload json.RawMessage}` — the read side of `WSMessage`. The `{type, payload}` wire format is byte-for-byte unchanged (no protocol break); only the in-memory representation differs. New typed `HeartbeatPayload` and `ChunkPayload` join the existing wire types (`RegisterPayload`, `FileAck`, ...).
- The claw bridge read loop (`claws/ws.go`) now dispatches through a per-connection registry `map[string]func(json.RawMessage) (wsHandlerFunc, error)`: each decoder unmarshals straight into the concrete type and returns a ready-to-run handler; the 400-line if/else chain became one small method per message type on `clawSession`. Malformed payloads for known types are **logged protocol errors** (previously silently dropped), and can never panic. Unknown message types are ignored, as before.
- The user WS read loop and the client `WatchMessages`/`WatchStream` decode the same way; the `interface{}` marshal round-trips are gone from every read path.
- `NormalizeAgentActivityPayload` now takes the raw wire bytes (`json.RawMessage`), so the persisted `activity:` format stores the payload verbatim instead of a re-marshalled map. Its `pkg/hub` alias test was updated for the new signature (forced by the signature change, no behavioral relaxation).

**Acceptance criteria from the spec:** no type assertion on `Payload` outside the decoder (verified — no `.Payload.(` assertions and no read-path `json.Marshal(msg.Payload)` remain in `pkg/hub`); the error→status mapping table is covered by `pkg/hub/httpserver/errors_test.go`; the decoder registry by `pkg/hub/claws/ws_decode_test.go`.

## Testing

- `go build ./...` — pass
- `go vet ./pkg/...` — pass
- `make test` (`go test -v ./...`) — all packages `ok`, zero failures
- `make test-factory` — pass
- `make test-parity` — pass
- `go test -race -count=1 ./pkg/hub/ ./pkg/hub/claws/ ./pkg/hub/httpserver/` — `claws` and `httpserver` pass; `pkg/hub` fails on `TestMidStreamDisconnectPreservesReceivedContent` with a DATA RACE in the disconnect flush (`HandleClawWS`'s deferred `StreamingBuf.Reset()` vs. the chunk handler). **This race is pre-existing: the identical failure reproduces on the base branch `feature/hub-ws-worker-pool`** (verified in a clean worktree of that branch); it is not introduced by this PR and should be fixed on the base branch of the stack.

### DIFF
--- a/pkg/hub/claw_handlers.go
+++ b/pkg/hub/claw_handlers.go
@@ -12,6 +12,7 @@ import (
 	"strings"
 
 	"github.com/elasticclaw/elasticclaw/pkg/hub/store"
+	"github.com/elasticclaw/elasticclaw/pkg/hub/httpserver"
 	"github.com/elasticclaw/elasticclaw/pkg/types"
 	"github.com/google/uuid"
 	"nhooyr.io/websocket"
@@ -502,7 +503,7 @@ func (s *Server) handleClawDetail(w http.ResponseWriter, r *http.Request) {
 
 	c, err := s.st().Claws().Get(r.Context(), clawID, tenantID)
 	if err == sql.ErrNoRows {
-		writeErr(w, http.StatusNotFound, "not_found", "not found")
+		httpserver.WriteDomainErr(w, types.ErrClawNotFound)
 		return
 	}
 	if err != nil {
--- a/pkg/hub/claw_handlers.go
+++ b/pkg/hub/claw_handlers.go
@@ -11,8 +11,8 @@ import (
 	"os"
 	"strings"
 
-	"github.com/elasticclaw/elasticclaw/pkg/hub/store"
 	"github.com/elasticclaw/elasticclaw/pkg/hub/httpserver"
+	"github.com/elasticclaw/elasticclaw/pkg/hub/store"
 	"github.com/elasticclaw/elasticclaw/pkg/types"
 	"github.com/google/uuid"
 	"nhooyr.io/websocket"
--- a/pkg/hub/claws/terminal.go
+++ b/pkg/hub/claws/terminal.go
@@ -14,6 +14,9 @@ import (
 
 	gossh "golang.org/x/crypto/ssh"
 	"nhooyr.io/websocket"
+
+	"github.com/elasticclaw/elasticclaw/pkg/hub/httpserver"
+	"github.com/elasticclaw/elasticclaw/pkg/types"
 )
 
 // HandleTerminal upgrades the request to a WebSocket and bridges it to an
@@ -26,7 +29,7 @@ func (s *Service) HandleTerminal(w http.ResponseWriter, r *http.Request) {
 	}
 	tenantID, err := s.tenantByToken(token)
 	if err != nil {
-		http.Error(w, "unauthorized", http.StatusUnauthorized)
+		httpserver.WriteDomainErr(w, types.ErrUnauthorized)
 		return
 	}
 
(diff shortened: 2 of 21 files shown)


pkg/hub/claws/terminal.go
```diff
@@ -14,6 +14,9 @@ import (
 
 	gossh "golang.org/x/crypto/ssh"
 	"nhooyr.io/websocket"
+
+	"github.com/elasticclaw/elasticclaw/pkg/hub/httpserver"
```
AnaBerg: manual review: This adds a `claws` service-package dependency on `pkg/hub/httpserver`, which conflicts with the phase-2 dependency rule that services should not know `httpserver` and should stay behind the HTTP boundary. The same pattern appears in `pkg/hub/checkpoints/checkpoints.go`; please keep the HTTP error-envelope mapping in the hub/httpserver layer or inject/return typed errors so the service packages only depend on `pkg/types` for domain errors.

pkg/hub/claw_handlers.go
```diff
@@ -12,6 +12,7 @@ import (
 	"strings"
 
 	"github.com/elasticclaw/elasticclaw/pkg/hub/store"
+	"github.com/elasticclaw/elasticclaw/pkg/hub/httpserver"
```
AnaBerg, line 14: manual review: `gofmt -l` reports this file, and also `pkg/hub/message_handlers.go`, because the new import is not sorted with the existing local imports. Please run gofmt on the touched Go files so a formatting check does not fail this PR.

AnaBerg (commented):
manual review: 2 inline findings.

PR #474, opened 2026-07-09
refactor(hub): remove deprecated ?token= fallback, finish error envelope migration

AnaBerg:
## Motivation

Closes the transitions opened earlier: the deprecated `?token=` query fallback (Phase 0.3) is removed, and the remaining handlers migrate to the unified JSON error envelope (Phase 0.2).

Concretely, on the base branch:

- `withAuth` and `withConfigMutationAuth` (pkg/hub/auth.go) still fell back to `r.URL.Query().Get("token")` when the `Authorization` header was empty, and `claws/terminal.go` even preferred the query param. Tokens in URLs leak into access logs, proxy logs, and browser history.
- 224 handler call sites across `pkg/hub` still wrote plain-text errors via `http.Error` instead of the `{"error", "code"}` envelope that `httpserver.WriteErr` produces (the `.golangci.yml` depguard notes explicitly tolerate services importing `httpserver` for these helpers until this migration lands).

## Dependencies

This PR is STACKED on branch `feature/domain-errors-typed-ws` — depends on #473 — review and merge that first.

Logically also depends on `feature/auth-token-hardening` (Phase 0.3) and `feature/recovery-middleware` (Phase 0.2) being merged first — this PR removes the transition paths they introduced. Those changes are **not** visible on the base branch of this stack, so the removal was implemented against the current code: the base still had the raw pre-deprecation `?token=` fallback (no single-use ticket mechanism, no deprecation warning), and `web/lib/api.ts` still builds WS/terminal URLs with `?token=`. Browser WebSocket clients cannot set headers, so this PR must land **after** the single-use-ticket work (hub tickets + the web change to use them); merging it standalone would break the web UI's WS and terminal connections. Expect small merge conflicts in `auth.go`/`terminal.go` where the ticket branch adds `?ticket=` handling next to the code removed here.

## What changed

- `pkg/hub/auth.go`: `withAuth` and `withConfigMutationAuth` accept the `Authorization` header (plus the web-session header for config mutations) only; the query-token branches are gone.
- `pkg/hub/claws/terminal.go`: the terminal WS handler authenticates via the `Authorization` header only.
- `pkg/hub/client.go`: the Go client sends `Authorization: Bearer <token>` on the `/api/ws` upgrade request instead of appending `?token=`.
- Tests updated: `hub_test.go` now asserts query tokens are rejected with 401; `message_handlers_test.go` dials the user WS with a Bearer header.
- Error envelope: all remaining `http.Error` calls in hub handlers (24 files, 224 call sites — auth, workspaces, templates, MCP, secrets, analytics, troubleshoot, claws files/terminal, checkpoints, volumes, cron/factory workflows, integrations webhooks, settings doctor, httpserver router) were mechanically rewritten to `writeErr`/`httpserver.WriteErr` with a stable status-derived `code` (`bad_request`, `unauthorized`, `not_found`, `internal`, …). Messages and HTTP statuses are unchanged; `factorytest` mocks (which emulate external services such as GitHub/Shortcut) keep plain text.

## Testing

- `go build ./...` — pass.
- `go test ./...` (equivalent to `make test`, minus `-v`) — all packages pass, including `pkg/hub` and `test/e2e`.
- `go test ./pkg/...` after the final commit — pass.
- `golangci-lint run ./pkg/hub/...` — 115 issues, identical to the base-branch baseline (all pre-existing; this diff adds none).

### DIFF
--- a/pkg/hub/auth.go
+++ b/pkg/hub/auth.go
@@ -87,7 +87,7 @@ func (s *Server) withConfigMutationAuth(next http.HandlerFunc) http.HandlerFunc 
 		if sessionSecret != "" {
 			if payload, ok := verifyGitHubSession(sessionSecret, token); ok {
 				if !isAccessAdmin(accessCfg, payload.Login) {
-					http.Error(w, "forbidden", http.StatusForbidden)
+					writeErr(w, http.StatusForbidden, "forbidden", "forbidden")
 					return
 				}
 				ctx := context.WithValue(r.Context(), ctxGitHubLoginKey{}, payload.Login)
@@ -97,7 +97,7 @@ func (s *Server) withConfigMutationAuth(next http.HandlerFunc) http.HandlerFunc 
 			}
 		}
 
-		http.Error(w, "unauthorized", http.StatusUnauthorized)
+		writeErr(w, http.StatusUnauthorized, "unauthorized", "unauthorized")
 	}
 }
 
@@ -187,7 +187,7 @@ func (s *Server) withWebAuth(next http.HandlerFunc) http.HandlerFunc {
 			token = r.Header.Get(webSessionHeader)
 		}
 		if token == "" {
-			http.Error(w, "unauthorized", http.StatusUnauthorized)
+			writeErr(w, http.StatusUnauthorized, "unauthorized", "unauthorized")
 			return
 		}
 		s.cfgMu.RLock()
@@ -212,7 +212,7 @@ func (s *Server) withWebAuth(next http.HandlerFunc) http.HandlerFunc {
 			}
 		}
 
-		http.Error(w, "unauthorized", http.StatusUnauthorized)
+		writeErr(w, http.StatusUnauthorized, "unauthorized", "unauthorized")
 	}
 }
 
@@ -235,7 +235,7 @@ func (s *Server) withWebAdminAuth(next http.HandlerFunc) http.HandlerFunc {
 			token = r.Header.Get(webSessionHeader)
 		}
 		if token == "" {
-			http.Error(w, "unauthorized", http.StatusUnauthorized)
+			writeErr(w, http.StatusUnauthorized, "unauthorized", "unauthorized")
 			return
 		}
 		s.cfgMu.RLock()
@@ -256,7 +256,7 @@ func (s *Server) withWebAdminAuth(next http.HandlerFunc) http.HandlerFunc {
 		if sessionSecret != "" {
 			if payload, ok := verifyGitHubSession(sessionSecret, token); ok {
 				if !isAccessAdmin(accessCfg, payload.Login) {
-					http.Error(w, "forbidden", http.StatusForbidden)
+					writeErr(w, http.StatusForbidden, "forbidden", "forbidden")
 					return
 				}
 				ctx := context.WithValue(r.Context(), ctxGitHubLoginKey{}, payload.Login)
@@ -266,7 +266,7 @@ func (s *Server) withWebAdminAuth(next http.HandlerFunc) http.HandlerFunc {
 			}
 		}
 
-		http.Error(w, "unauthorized", http.StatusUnauthorized)
+		writeErr(w, http.StatusUnauthorized, "unauthorized", "unauthorized")
 	}
 }
 
@@ -340,19 +340,19 @@ func (s *Server) handleAuthConfig(w http.ResponseWriter, r *http.Request) {
 
 func (s *Server) handleLogin(w http.ResponseWriter, r *http.Request) {
 	if r.Method != http.MethodPost {
-		http.Error(w, "method not allowed", http.StatusMethodNotAllowed)
+		writeErr(w, http.StatusMethodNotAllowed, "method_not_allowed", "method not allowed")
 		return
 	}
 	var body struct {
 		Token string `json:"token"`
 	}
 	if err := json.NewDecoder(r.Body).Decode(&body); err != nil || body.Token == "" {
-		http.Error(w, "invalid request", http.StatusBadRequest)
+		writeErr(w, http.StatusBadRequest, "bad_request", "invalid request")
 		return
 	}
 	tenantID, err := s.tenantByToken(body.Token)
 	if err != nil {
-		http.Error(w, "invalid token", http.StatusUnauthorized)
+		writeErr(w, http.StatusUnauthorized, "unauthorized", "invalid token")
 		return
 	}
 	jsonOK(w, map[string]string{"tenant_id": tenantID, "token": body.Token})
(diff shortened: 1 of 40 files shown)


pkg/hub/checkpoints/volumes.go
```diff
@@ -401,29 +401,29 @@ func (s *Service) HandleVolumeArchive(w http.ResponseWriter, r *http.Request) {
...
-		http.Error(w, "method not allowed", http.StatusMethodNotAllowed)
+		httpserver.WriteErr(w, http.StatusMethodNotAllowed, "method_not_allowed", "method not allowed")
 	}
 }
 
 func (s *Service) handleVolumeArchiveGet(w http.ResponseWriter, r *http.Request, leaseID, clawID, accessToken string) {
 	manifestDigest, err := s.volumeLeaseManifest(r.Context(), leaseID, clawID, accessToken)
 	if err != nil {
-		http.Error(w, err.Error(), http.StatusNotFound)
+		httpserver.WriteErr(w, http.StatusNotFound, "not_found", err.Error())
```
AnaBerg: CodeRabbit finding (minor): `volumeLeaseManifest` can fail for reasons other than a missing lease, but this path maps every error to 404 and returns `err.Error()` to the client. Please distinguish `sql.ErrNoRows` or an explicit not-found sentinel from backend/DB failures, return a safe 404 only for the former, and use a 500/internal error without leaking raw backend text for the latter.

pkg/hub/checkpoints/volumes.go
```diff
@@ -434,25 +434,25 @@ func (s *Service) handleVolumeArchiveGet(w http.ResponseWriter, r *http.Request,
 func (s *Service) handleVolumeArchivePut(w http.ResponseWriter, r *http.Request, leaseID, clawID, accessToken string) {
 	var repo, tag, mode, attachedDigest string
 	if err := s.deps.DB.QueryRow(`SELECT repo, tag, mode, manifest_digest FROM volume_leases WHERE id=? AND claw_id=? AND access_token=? AND released_at IS NULL AND expires_at > ?`, leaseID, clawID, accessToken, time.Now().UTC()).Scan(&repo, &tag, &mode, &attachedDigest); err != nil {
-		http.Error(w, "active lease not found", http.StatusNotFound)
+		httpserver.WriteErr(w, http.StatusNotFound, "not_found", "active lease not found")
```
AnaBerg: CodeRabbit finding (minor): this `Scan` error path treats every DB failure as a missing active lease. Please check `errors.Is(err, sql.ErrNoRows)` and keep the current 404 only for no rows; route other DB/backend errors through the handler's internal-error response so operational failures are not misreported as not found.

pkg/hub/checkpoints/volumes.go
```diff
@@ -434,25 +434,25 @@ func (s *Service) handleVolumeArchiveGet(w http.ResponseWriter, r *http.Request,
...
 	}
 	if mode != volumeModeRW {
-		http.Error(w, "volume is read-only", http.StatusForbidden)
+		httpserver.WriteErr(w, http.StatusForbidden, "forbidden", "volume is read-only")
 		return
 	}
 	currentDigest, err := s.artifacts().ResolveRef(r.Context(), repo, tag)
 	if err != nil {
-		http.Error(w, "resolve volume ref: "+err.Error(), http.StatusConflict)
+		httpserver.WriteErr(w, http.StatusConflict, "conflict", "resolve volume ref: "+err.Error())
```
AnaBerg: CodeRabbit finding (minor): `ResolveRef` failures are currently returned as `409 conflict`, which conflates transient artifact/registry lookup errors with the later digest-mismatch conflict. Please return an internal/server error for lookup failures while keeping `409 conflict` for the `currentDigest != attachedDigest` case.

pkg/hub/auth.go
```diff
@@ -17,17 +17,18 @@ import (
 
 func (s *Server) withAuth(next http.HandlerFunc) http.HandlerFunc {
 	return func(w http.ResponseWriter, r *http.Request) {
+		// Authorization header only. The deprecated ?token= query fallback was
+		// removed (Phase 2.6): tokens in URLs leak into access logs, proxy logs,
+		// and browser history. Endpoints that cannot set headers (WS upgrades,
+		// <img src>) authenticate with a single-use ticket instead.
```
AnaBerg, line 23: docs/re-arch spec-alignment finding: Phase 0.3 says browser-only paths (`/api/ws`, `/api/terminal/{id}`, and `/api/files/view/...`) should move from `?token=` to a single-use ticket before the deprecated query-token fallback is removed in Phase 2.6. This PR removes query-token auth from `withAuth`/terminal, but I do not see a `/api/auth/ticket` issuer/redeemer in the server, and `web/lib/api.ts` still builds `?token=` URLs for WS, terminal, and file previews. As-is, browser WS and `` file preview flows will start returning 401. Please add the ticket flow and update the frontend URLs, or keep the fallback until that ticket path lands.

AnaBerg (commented):
PR review via CodeRabbit plus a separate docs/re-arch spec-alignment pass. Inline comments label their source explicitly.

AnaBerg:
The single-use ticket implementation still bypasses per-user claw access rules on the terminal endpoint.

Tickets retain the GitHub login when issued, and `withAuthOrTicket` restores that login for the user WS/file paths. However, `redeemAuthTicket` deliberately returns only `tenantID`, and `claws.HandleTerminal` then authorizes solely through `SSHEndpoint(clawID, tenantID)`. An OAuth user restricted by tag filters can therefore open an SSH terminal to any claw in the same tenant if they know its ID, even though list/detail/message paths apply `canViewClaw`/`canModifyClaw`.

Please preserve the login through terminal ticket/header authentication and enforce the same per-claw authorization—preferably modify-level access—before looking up or dialing the SSH endpoint.

AnaBerg:
The ticket stores `githubLogin`, but the terminal bridge discards it: `redeemAuthTicket` returns only the tenant ID, and `HandleTerminal` authorizes solely by tenant ownership. Consequently any authenticated GitHub user who can mint a ticket can obtain an SSH shell for any claw in that tenant, even when tag-based access grants only view/no modify permission. Preserve the login through the `RedeemAuthTicket` dependency and run the same `canModifyClaw` authorization used by mutating endpoints before opening SSH.

AnaBerg:
The final error-envelope migration still leaves service packages importing `pkg/hub/httpserver`, despite Phase 2.1 requiring services not to know the HTTP transport and the depguard comment promising this exception would end with Phase 2.5. Please keep error-to-status/envelope mapping in the HTTP adapter (or move transport-neutral error data below both layers) and remove these upward imports before closing the compatibility phase.

PR #478, opened 2026-07-13
Agent logs in analytics drill-down

AnaBerg:
## What

Adds an **Agent logs** dialog to the task-run analytics drill-down (`RunDetailPanel`), with two tabs:

- **Actions** — the agent activity stream (tool calls, commands, paths, errors) for the selected attempt, fetched from the existing `GET /api/messages/{clawId}/activity` endpoint, with an attempt selector for multi-attempt runs and "Load older" pagination.
- **Output** — pipeline stage stdout/stderr with exit-code badges, via a new `GET /api/analytics/runs/{runId}/outputs` endpoint that exposes `pipeline_outputs` for the run's claws (run claw + each attempt's claw), tenant-scoped.

## Why

The drill-down previously showed only summary/attempts/events/PRs; there was no way to see what an agent actually did during a run without leaving the analytics screen.

## Security fix included

The analytics endpoints only enforced tenant scoping, while the messages endpoints enforce the tag-based per-claw ACL (`canViewClaw`). Linking analytics to logs made that gap visible, so this PR mirrors the ACL on the analytics run detail and all `/runs/{runId}/*` subresources (403/404 for restricted GitHub-OAuth callers) and filters restricted runs out of the list.

Notes for reviewers:
- The ACL check short-circuits when no view restrictions are configured — important because claw rows are hard-deleted on run completion while `task_run_summaries` persist; without the short-circuit, OAuth users would lose all historical analytics.
- List filtering batch-fetches claw tags per page (single `IN (...)` query) to avoid N+1.
- Known follow-up (intentionally out of scope): `GET /api/analytics/summary` and `/filter-options` aggregates are still tenant-wide, so a tag-restricted user can see counts/breakdowns covering runs they cannot open.

## Testing

- New Go tests for the `/outputs` endpoint and ACL enforcement (including deleted-claw behavior, restricted and unrestricted): `go test ./pkg/hub/` passes.
- `go build ./...`, `tsc --noEmit`, eslint on touched files, and `next build` all pass.

### DIFF
--- a/pkg/hub/task_run_analytics_api.go
+++ b/pkg/hub/task_run_analytics_api.go
@@ -10,6 +10,8 @@ import (
 	"strconv"
 	"strings"
 	"time"
+
+	"github.com/elasticclaw/elasticclaw/pkg/types"
 )
 
 const taskRunAnalyticsDefaultLimit = 50
@@ -72,6 +74,10 @@ type taskRunAnalyticsEventsResponse struct {
 
 type taskRunAnalyticsPRsResponse struct {
 	PRs []taskRunAnalyticsPRView `json:"prs"`
+}
+
+type taskRunAnalyticsOutputsResponse struct {
+	Outputs []taskRunAnalyticsOutputView `json:"outputs"`
 }
 
 type taskRunAnalyticsFilterOptionsResponse struct {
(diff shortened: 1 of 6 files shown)


AnaBerg:
Both Greptile findings addressed in e457527:

- **Unsanitized URL in href**: activity URLs are now only rendered as links when the parsed scheme is `http:`/`https:` (otherwise shown as plain text).
- **Attempt selector on Output tab**: the tabs are now controlled and the attempt selector only renders on the Actions tab, where it applies.

PR #484, opened 2026-07-15
fix: graceful shutdown, panic recovery, sqlite busy_timeout, pr_conditions ordering

AnaBerg:
## Summary

Agent runs could get stuck forever in non-terminal states: an unhandled panic in a run-owning goroutine silently killed background work with no recovery, a hard process exit during shutdown could interrupt in-flight requests and race with DB close, SQLite writers could contend without a bounded wait, and a crash between marking `pr_conditions_fired` and firing the pipeline transition could permanently strand a PR trigger. This change hardens the hub server against that class of liveness bug: goroutines recover and log instead of taking the process down, shutdown drains cleanly before closing resources, SQLite waits on `busy_timeout` instead of failing immediately on lock contention, and `pr_conditions` are only marked fired after the transition succeeds.

## Changes

- Graceful shutdown: `Server.Run` wraps `http.ListenAndServe` in an `http.Server` wired to `signal.NotifyContext(SIGINT, SIGTERM)`, calls `srv.Shutdown` with a 10s timeout, and only closes the DB after shutdown completes; `http.ErrServerClosed` is treated as a clean exit (`pkg/hub/server.go`).
- Panic recovery: added `Server.safeGo(name, fn)` (defer/recover + stack log) and applied it to all run-owning goroutines — pipeline judge/gate auto-transition, transition/message-trigger/done-signal goroutines, async webhook handlers (Linear, GitHub issues, Jira, Shortcut, GitHub webhook, external webhook), the integration poller and each tick, and cron job execution via `cron.WithChain(cron.Recover(...))`.
- SQLite `busy_timeout`: confirmed `busy_timeout(5000)` is set in the DSN pragma (`db.go`) and added a regression test asserting `PRAGMA busy_timeout` returns 5000.
- `pr_conditions` ordering: `pr_watcher.go` now runs the pipeline transition first via a new `firePRConditions` helper and only sets `pr_conditions_fired=1` after the transition claim succeeds, so a crash or failed transition leaves the trigger eligible for the next poll.
- Added/extended tests: `server_shutdown_test.go`, `safe_go_test.go`, `pr_watcher_test.go`, `db_test.go`.

## Testing

- `go build ./...` and `go test ./pkg/hub/...` passing.

## Review

Review verdict: needs_fixes with 4 finding(s). Blocking findings were fixed afterwards: Fixed both blocking review findings via Codex (gpt-5.6). (1) pkg/hub/server.go: added a shutdownDone channel closed after srv.Shutdown returns; Run() now waits on it in the ErrServerClosed branch before calling s.db.Close(), so the process no longer exits mid-drain and in-flight requests are no longer hard-killed nor racing with DB close. (2) pkg/hub/safe_go_test.go: replaced the raw bytes.Buffer with a mutex-guarded lockedBuffer (locking both Write and String) in TestSafeGoRecoversAndLogsPanic, eliminating the data race between safeGo's recover-handler log write and the test's polling read. Verified: go build ./... succeeds; go test ./pkg/hub/... passes (pkg/hub, artifact, factorytest, pipeline all ok); go test -race -run TestSafeGoRecoversAndLogsPanic ./pkg/hub and go test -race -run TestRunShutsDownCleanly ./pkg/hub both pass. Committed as c035a64.

### DIFF
--- a/pkg/hub/cron_scheduler.go
+++ b/pkg/hub/cron_scheduler.go
@@ -63,7 +63,7 @@ func (cs *cronScheduler) start() error {
 	defer cs.mu.Unlock()
 
 	// Create cron with second-level precision (optional) and timezone support
-	cs.cron = cron.New(cron.WithSeconds())
+	cs.cron = cron.New(cron.WithSeconds(), cron.WithChain(cron.Recover(cron.DefaultLogger)))
 
 	// Load all workspaces and register cron workflows
 	if err := cs.reloadWorkflows(); err != nil {
--- a/pkg/hub/db_test.go
+++ b/pkg/hub/db_test.go
@@ -8,6 +8,22 @@ import (
 	"time"
 )
 
+func TestOpenDBSetsBusyTimeout(t *testing.T) {
+	db, err := openDB(filepath.Join(t.TempDir(), "hub.db"))
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer db.Close()
+
+	var timeout int
+	if err := db.QueryRow(`PRAGMA busy_timeout`).Scan(&timeout); err != nil {
+		t.Fatal(err)
+	}
+	if timeout != 5000 {
+		t.Fatalf("busy_timeout = %d, want 5000", timeout)
+	}
+}
+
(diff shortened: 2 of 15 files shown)


AnaBerg:

PR #485, opened 2026-07-15
fix: make pipeline on_enter/gate failures terminal and handle issue-less [DONE]

AnaBerg:
## Summary

Fixes a class of liveness bugs where agent runs could get stuck forever in non-terminal states. Several code paths silently swallowed failures instead of terminating the run: a failing `on_enter` action would drop remaining actions (including injects) without stopping the claw/workflow run; a gate verdict of `error` with no declared route would leave the pipeline stalled instead of failing; and `[DONE]` signals from issue-less workflow claws were discarded instead of driving the claw to `idle` or completing the run.

## Changes

- `runOnEnter` now returns a descriptive error; non-terminal stage `on_enter` failures call `stopAgentWithReason` (claw -> `error`, workflow run -> `failed`) instead of silently succeeding while dropping remaining actions.
- `autoTransitionAfterGate` calls `stopAgentWithReason` when a gate verdict is `error` and no stage declares an `error` gate_result route; pipelines with a declared error route still transition normally.
- `initializePipelineEntryIfNeeded` now only reports the inject as delivered when it was actually sent (defense in depth alongside the `on_enter` fix).
- `handleClawDoneSignal` (linear.go) no longer discards `[DONE]` signals from issue-less workflow claws: it validates/stores PR URLs, sets the claw to `idle` if PRs are pending, otherwise runs the terminal teardown and calls the cron scheduler's `finishRunByClawID("completed")`.
- Added tests in `pipeline_runner_test.go` (non-terminal on_enter failure, gate error without route, gate error with declared route) and `done_signal_test.go` (issue-less `[DONE]` with/without PRs).

## Testing

- `go build` and `go test ./pkg/hub/...` passing.

## Review

Review verdict: needs_fixes with 5 finding(s). Blocking findings were fixed afterwards: Fixed both blocking review findings via Codex CLI (gpt-5.6). (1) pkg/hub/pipeline_runner.go: added a routedRequiredGateError type so a required-gate fail/error whose verdict has a declared gate_result route no longer triggers stopAgentWithReason in transitionResolvedPipelineStageWithContext — it now lets autoTransitionAfterGate's goroutine own the declared route, restoring the gate retry-loop pattern. Unrouted required-gate failures still terminate the run (liveness preserved). (2) pkg/hub/linear.go: handleClawDoneSignal now checks for an associated running workflow_runs row before calling completeIssueLessDoneClaw on an issue-less claw; interactive claws with no workflow run are a no-op again, as before. Added test coverage in pkg/hub/done_signal_test.go (new TestHandleClawDoneSignal_IssueLessInteractiveClawIsNoOp, plus a workflow_runs row inserted into the existing issue-less PR test).

### DIFF
--- a/pkg/hub/done_signal_test.go
+++ b/pkg/hub/done_signal_test.go
@@ -375,3 +375,52 @@ func TestHandleClawDoneSignal_AllowedWhenNoFailedRequiredGate(t *testing.T) {
 		t.Fatalf("expected 0 PRs stored (no GH App), got %d", count)
 	}
 }
+
+func TestHandleClawDoneSignal_IssueLessWorkflowCompletesWithoutPR(t *testing.T) {
+	s, db := NewTestServerWithConfig(t, &types.HubConfig{Token: "test-token"}, "", "", "")
+	s.cronScheduler = newCronScheduler(s)
+	const clawID = "claw-issue-less-done"
+	_, err := db.Exec(`INSERT INTO claws(id, tenant_id, name, template, status, created_at) VALUES(?,?,?,?,?,datetime('now'))`, clawID, "test-tenant-id", "workflow claw", "base", "connected")
+	if err != nil {
+		t.Fatal(err)
+	}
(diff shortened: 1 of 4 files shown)


AnaBerg:

PR #486, opened 2026-07-15
fix: add boot reconciliation and stale-state reaper

AnaBerg:
## Summary

Agent runs could get stuck forever in non-terminal states (e.g. claws stuck `provisioning`/`starting`, orphaned `running` workflow runs, stale claimed factory triggers) with nothing to sweep them out. This fixes that class of liveness bug by adding a boot-time reconciler plus a periodic reaper that detects and clears stuck state, and by making failure paths correctly promote pending claws and retry provider-lost VMs.

## Changes

- Add `LivenessConfig` (`pkg/types/template.go`) with `enabled`, `offline_grace`, `provisioning_max_age`, `claim_ttl`, `reaper_interval`.
- Add `pkg/hub/reaper.go`:
  - `reconcileOnBoot`: sweeps stuck provisioning/starting claws, orphaned running workflow_runs, and stale claimed factory_triggers, creating checkpoints, then promotes pending claws. Runs synchronously in `NewServer` before pollers/cron/PR watcher start.
  - `runReaper`: ticker-based, panic-recovering background loop with in-memory `firstSeen` tracking, an action cap (~20/tick), enforcing `offline_grace`, `provisioning_max_age`, `claim_ttl`, and `task_run` `timeout_at`.
  - Both paths funnel through the existing `stopAgentWithReason`/`promotePendingClaws`.
- `stopAgentWithReason` now calls `promotePendingClaws` at the end (`pkg/hub/pipeline_runner.go`).
- `syncReplicatedVMs`' VM-404 branch now calls `stopAgentWithReason(..., "replicated VM no longer exists", true)` instead of just marking the claw offline (`pkg/hub/server.go`).
- Add `pkg/hub/reaper_test.go` covering boot reconciliation, reaper grace/timeout behavior with an injectable `nowFunc`, `firstSeen` clearing on reconnect, and pending-claw promotion.

## Testing

- `go build ./...` passing
- `go test ./pkg/hub/...` passing

## Review

Review verdict: needs_fixes with 7 finding(s). Blocking findings were fixed afterwards: Fixed both blocking findings via Codex (gpt-5.6): (1) pkg/hub/reaper.go now ages provisioning/starting claws from when the reaper first observes them in that status, instead of claws.created_at, so claws promoted from pending or mid-retry re-provisions are no longer killed on the first tick after starting; added TestReaperProvisioningMaxAgeStartsWhenProvisioningIsObserved covering both the long-pending-then-promoted case (not killed) and the genuinely-stuck case (killed after provisioning_max_age). (2) pkg/hub/server.go:4986 restores a "Provider VM lost: replicated VM no longer exists" reason string (keeping the "Provider VM lost" signal) so classifyAgentFailure still resolves the replicated-VM-404 path to the retryable agentFailureProviderLost instead of falling through to non-retryable agentFailureUnknown; claw_retry_test.go updated accordingly. Verified with `go build ./...` and `go test ./pkg/hub/...` outside the Codex sandbox (both pass; Codex's own sandbox run of the full suite was blocked by a loopback-listener restriction unrelated to these changes, so I reran it directly).

### DIFF
--- a/pkg/hub/claw_retry_test.go
+++ b/pkg/hub/claw_retry_test.go
@@ -312,7 +312,7 @@ func TestHealthEscalationThresholds(t *testing.T) {
 }
 
 func TestProviderLostClassificationAndFailureType(t *testing.T) {
-	failure := classifyAgentFailure("Provider VM lost: HTTP 404 not found")
+	failure := classifyAgentFailure("Provider VM lost: replicated VM no longer exists")
 	if failure.Kind != agentFailureProviderLost {
 		t.Fatalf("kind=%q, want %q", failure.Kind, agentFailureProviderLost)
 	}
--- a/pkg/hub/pipeline_runner.go
+++ b/pkg/hub/pipeline_runner.go
@@ -1782,9 +1782,11 @@ func (s *Server) stopAgentWithReason(clawID, reason string, skipVMTerminate bool
 	}, time.Sleep)
 	switch disposition {
 	case clawRetryScheduled, clawRetryAlreadyHandled:
+		s.promotePendingClaws()
 		return
 	}
 	s.stopAgentTerminalWithReason(clawID, reason, skipVMTerminate)
+	s.promotePendingClaws()
 }
 
 func (s *Server) stopAgentTerminalWithReason(clawID, reason string, skipVMTerminate bool) {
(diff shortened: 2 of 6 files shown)


AnaBerg:

PR #550, opened 2026-07-24
feat: add workflow run history to UI and CLI

xavpaice:
Adds visibility into cron/manual workflow execution history.

### Changes
- **Web UI:** New "Run history" button on Settings → Workflows rows. Opens a dialog listing recent runs with status, trigger type, start/finish times, full result text, and agent ID. Includes a link to full task-run analytics for the workflow.
- **CLI:** New `elasticclaw workflow runs <name> --workspace <workspace> --limit <n>` command with JSON output support.
- **Backend:** Uses the existing `/api/workspaces/{ws}/workflows/{wf}/cron/runs` endpoint.

### UI polish
- Content area scrolls vertically.
- Result text wraps and preserves line breaks instead of overflowing into adjacent columns.
- Analytics link uses the router and closes the dialog before navigating.

### Commits
- `3f4b5aaf` feat(web): add workflow run history dialog to workflows settings page
- `7d636205` feat(cli): add workflow runs command to list execution history
- `1112bab0` fix(web): scroll run history dialog and wrap result text legibly
- `f376a95a` fix(web): preserve line breaks and show full result text in run history

Tested and deployed on a remote VM; the UI and CLI both surface the failed cron run results correctly.

image

### DIFF
--- a/cmd/workflow.go
+++ b/cmd/workflow.go
@@ -26,6 +26,7 @@ func WorkflowCmd() *cobra.Command {
 	cmd.AddCommand(workflowShowCmd())
 	cmd.AddCommand(workflowPushCmd())
 	cmd.AddCommand(workflowTriggerCmd())
+	cmd.AddCommand(workflowRunsCmd())
 	return cmd
 }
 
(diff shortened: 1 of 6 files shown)


xavpaice:
One thing to note before merging: the run history shown here only covers **cron and manual-triggered workflows**.

Workflows triggered by issue trackers (GitHub Issues, Linear, Shortcut, Jira) call `createClawFromWorkflowWithOptions` directly and do not write to the `workflow_runs` table. To make this dialog and CLI command universal, we would need to add `workflow_runs` inserts in those paths too — which means more rows in the database for every issue-tracker-driven agent run.

Given the earlier concern about retention and table growth, I think we should decide whether we want that expansion before adding it. The current change is intentionally scoped to cron/manual history only.

marccampbell (approved):

PR #551, opened 2026-07-24
fix(hub): stop 12-min watchdog nag flood; deliver nudge mid-turn; auto-resume after gateway restart

AnaBerg:
## Incident (NEXT-490)

On 2026-07-24, a long-running agent turn triggered the 12-minute watchdog repeatedly: **35 duplicate "still working?" nags were queued between 12:01 and 12:18 UTC**. When the gateway restarted mid-incident, the entire queue drained into a freshly reset session, and the claw then sat **stale for ~50 minutes** until it was manually resumed.

## Changes

1. **Turn-scoped watchdog flags** — `streamingTimeoutSent` is scoped to the current turn and no longer re-arms while the same turn is still streaming, so the watchdog fires at most once per turn instead of flooding.
2. **Stale-nag drop** — pending watchdog nags addressed to a turn that has already ended are dropped instead of being delivered into the next turn or a reset session. `sendStreamingNudge` re-checks busy state at entry and again after a failed status-channel write before falling back to the queue.
3. **Mid-turn nudge delivery over the status channel** — the bridge injects watchdog nudges into the in-flight OpenClaw turn via the WS status channel (`Channel:"status"`), so the agent sees the nudge while it is still working; if no status channel is connected, it falls back to the regular queue.
4. **Pending-queue dedup** — byte-identical pending queued messages are deduplicated (per role), so repeated watchdog/resume enqueues collapse to a single delivery.
5. **Auto-resume after gateway restart** — the hub detects a genuine gateway restart via the heartbeat restart counter and enqueues a resume automatically. The first heartbeat observation per claw is recorded as a baseline (no restart path), so a hub reboot cannot cause a spurious resume; `lastTurnFinishedAt` survives main-channel reconnects (copied from the old conn, or seeded from the newest `role='claw'` message when the bridge process itself restarted); restarts during bootstrap do not enqueue a resume.

## Test coverage

New/updated tests in `pkg/hub/watchdog_enforcement_test.go`, `pkg/hub/pr_watcher_test.go`, `cmd/claw-bridge/main_test.go`:

- `TestStreamingNudgeGoesOverStatusChannel` — nudge frame delivered over the WS status channel, UI row recorded, zero pending rows
- `TestStreamingNudgeFallsBackToQueueWithoutStatusChannel`
- `TestStreamingNudgeDroppedWhenTurnAlreadyEnded` — late fallback nudge does not leak into the next turn
- `TestRestartDuringBootstrapDoesNotEnqueueResume`
- `TestFirstHeartbeatAfterHubBootDoesNotAutoResume` — baseline observation, then genuine restart auto-resumes
- Interleaved-segments test asserts `streamingTimeoutSent` stays set across segment flushes (guards against re-arm regression)
- Dedup and mid-turn injection tests for the queue and bridge paths

## Verification

- `go build ./...` and `go vet ./...` clean
- `go test ./...` all packages ok (`pkg/hub` 24s)
- New/watchdog tests pass under `-race -count=3`
- `go test -tags integration ./pkg/hub/`: 6 `TestWorkflowHarness_*` failures are **pre-existing on main** — verified byte-identical failure set at merge-base 8256c62 in a clean worktree; this branch introduces no new integration failures

### DIFF
--- a/cmd/claw-bridge/main.go
+++ b/cmd/claw-bridge/main.go
@@ -1802,6 +1802,33 @@ func setActiveGatewaySession(gs *gatewaySession) {
 	gatewayProcessState.Unlock()
 }
 
+func activeGatewaySession() *gatewaySession {
+	gatewayProcessState.RLock()
+	defer gatewayProcessState.RUnlock()
+	return gatewayProcessState.session
+}
+func (gs *gatewaySession) turnInFlight() bool {
+	gs.infMu.RLock()
+	defer gs.infMu.RUnlock()
+	return gs.inFlight != nil
+}
+func injectGatewayNudge(ctx context.Context, text string) {
+	gs := activeGatewaySession()
+	if gs == nil || !gs.IsReady() {
+		log.Printf("[status] dropping nudge: gateway session not ready")
+		return
+	}
+	if !gs.turnInFlight() {
+		log.Printf("[status] dropping nudge: no turn in flight")
+		return
+	}
+	reqCtx, cancel := context.WithTimeout(ctx, 15*time.Second)
+	defer cancel()
+	if _, err := gs.sendReq(reqCtx, "sessions.send", map[string]string{"key": gs.getSessionKey(), "message": text}); err != nil {
+		log.Printf("[status] nudge injection failed: %v", err)
+	}
+}
+
 func (gs *gatewaySession) setReady() {
 	gs.readyMu.Lock()
 	gs.ready = true
@@ -3961,6 +3988,14 @@ func runStatusChannel(ctx context.Context, wsURL, clawID, clawName, templateName
 			}
 			if msg.Type == "status_ping" {
 				_ = wsjson.Write(ctx, conn, hubMsg{Type: "status_pong"})
+			}
+			if msg.Type == "nudge" {
+				var p struct {
+					Content string `json:"content"`
+				}
+				if err := json.Unmarshal(msg.Payload, &p); err == nil && p.Content != "" {
+					go injectGatewayNudge(ctx, p.Content)
+				}
 			}
 		}
 		pingCancel() // stop the ping goroutine before closing the connection
--- a/cmd/claw-bridge/main.go
+++ b/cmd/claw-bridge/main.go
@@ -1807,25 +1807,76 @@ func activeGatewaySession() *gatewaySession {
 	defer gatewayProcessState.RUnlock()
 	return gatewayProcessState.session
 }
-func (gs *gatewaySession) turnInFlight() bool {
-	gs.infMu.RLock()
-	defer gs.infMu.RUnlock()
-	return gs.inFlight != nil
-}
 func injectGatewayNudge(ctx context.Context, text string) {
 	gs := activeGatewaySession()
 	if gs == nil || !gs.IsReady() {
 		log.Printf("[status] dropping nudge: gateway session not ready")
 		return
 	}
-	if !gs.turnInFlight() {
+	reqCtx, cancel := context.WithTimeout(ctx, 15*time.Second)
+	defer cancel()
+	sent, err := gs.sendTurnScopedReq(reqCtx, "sessions.send", map[string]string{"key": gs.getSessionKey(), "message": text})
+	if !sent {
 		log.Printf("[status] dropping nudge: no turn in flight")
 		return
 	}
-	reqCtx, cancel := context.WithTimeout(ctx, 15*time.Second)
-	defer cancel()
-	if _, err := gs.sendReq(reqCtx, "sessions.send", map[string]string{"key": gs.getSessionKey(), "message": text}); err != nil {
+	if err != nil {
 		log.Printf("[status] nudge injection failed: %v", err)
+	}
+}
+
+// sendTurnScopedReq dispatches a request only while a turn is in flight,
+// holding infMu across the check AND the WebSocket write so the readLoop
+// cannot mark the turn finished in between (closing the turnInFlight/send
+// TOCTOU). The response is awaited without the lock — the readLoop needs
+// infMu.Lock to clear inFlight, so holding it across the await would
+// deadlock. Returns sent=false when no turn was in flight at write time.
+func (gs *gatewaySession) sendTurnScopedReq(ctx context.Context, method string, params interface{}) (bool, error) {
+	paramsJSON, _ := json.Marshal(params)
+	reqID := randomID()
+	ch := make(chan gwFrame, 1)
+
+	gs.pendMu.Lock()
+	gs.pending[reqID] = ch
+	gs.pendMu.Unlock()
+	cleanup := func() {
+		gs.pendMu.Lock()
+		delete(gs.pending, reqID)
+		gs.pendMu.Unlock()
+	}
+
+	gs.infMu.RLock()
+	if gs.inFlight == nil {
+		gs.infMu.RUnlock()
+		cleanup()
+		return false, nil
+	}
+	conn := gs.currentConn()
+	// Short write deadline: the RLock stalls readLoop writers (turn-end
+	// bookkeeping) until the write returns, so don't hold it for the full
+	// request timeout against a wedged socket.
+	writeCtx, writeCancel := context.WithTimeout(ctx, 5*time.Second)
+	err := wsjson.Write(writeCtx, conn, gwFrame{Type: "req", ID: reqID, Method: method, Params: paramsJSON})
+	writeCancel()
+	gs.infMu.RUnlock()
+	if err != nil {
+		cleanup()
+		return true, fmt.Errorf("%s write: %w", method, err)
+	}
+
+	select {
+	case resp := <-ch:
+		if !resp.OK {
+			msg := "unknown"
+			if resp.Error != nil {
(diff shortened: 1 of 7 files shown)


pkg/hub/server.go
```diff
@@ -2358,13 +2383,29 @@ func (s *Server) handleClawWS(w http.ResponseWriter, r *http.Request) {
...
+						} else if hb.RestartCount != lastRestartCount {
+							// Any change signals a restart: an increase is an in-process
+							// gateway restart; a decrease means the bridge process itself
+							// was relaunched and its counter reset.
 							log.Printf("[heartbeat] %s (%s): agent process restarted (restart_count=%d)", rp.Name, clawID[:8], hb.RestartCount)
 							s.gatewayRestartCounts[clawID] = hb.RestartCount
+							if s.autoResumeRestartCounts == nil {
+								s.autoResumeRestartCounts = make(map[string]int)
+							}
+							turnOpenOrRecent := !cc.streamingStartedAt.IsZero() || cc.awaitingResponse ||
```
AnaBerg, line 2403: **Critical — auto-resume dedup can't tell "never resumed" from "resumed at count 0".**

```go
if turnOpenOrRecent && s.autoResumeRestartCounts[clawID] != hb.RestartCount {
```

The map read returns the zero value `0` for an absent key. When the bridge process itself is relaunched and its `restart_count` resets to `0`, this compares `0 != 0` → false, so `shouldAutoResume` stays `false` and **no resume is ever enqueued**.

That's exactly the case the comment a few lines above says is handled ("a decrease means the bridge process itself was relaunched and its counter reset"). A claw whose bridge restarts mid-turn silently loses its turn.

Fix: use an existence-aware lookup.

```go
lastResumedCount, resumeRecorded := s.autoResumeRestartCounts[clawID]
if turnOpenOrRecent && (!resumeRecorded || lastResumedCount != hb.RestartCount) {
    s.autoResumeRestartCounts[clawID] = hb.RestartCount
    shouldAutoResume = true
}
```

Dedup for repeated restart counts (including repeated `0`) is preserved.

pkg/hub/pr_watcher.go
```diff
@@ -1049,6 +1049,12 @@ func (s *Server) injectMessage(clawID, content, role string) {
 	if err := s.db.QueryRow(`SELECT tenant_id FROM claws WHERE id=?`, clawID).Scan(&tenantID); err != nil {
 		return
 	}
+	var pendingDupes int
+	_ = s.db.QueryRow(`SELECT COUNT(1) FROM messages WHERE claw_id=? AND role=? AND content=? AND delivered_at IS NULL`, clawID, role, content).Scan(&pendingDupes)
```
AnaBerg: **Warning — discarded `Scan` error makes the dedup check fail open.**

```go
_ = s.db.QueryRow(`SELECT COUNT(1) FROM messages ...`).Scan(&pendingDupes)
```

If the query fails for any reason, `pendingDupes` stays `0` and we fall through to the injection path as though no duplicate existed — which is the opposite of what this guard is for. Given the enclosing function returns silently on the `tenant_id` query error just above, the consistent move is to fail closed (or at minimum log) rather than swallow it:

```go
if err := s.db.QueryRow(...).Scan(&pendingDupes); err != nil {
    log.Printf("[pr-watcher] dedup check failed for claw %s: %v", shortID(clawID), err)
    return
}
```

cmd/claw-bridge/main.go
```diff
@@ -1802,6 +1802,33 @@ func setActiveGatewaySession(gs *gatewaySession) {
...
+	defer gs.infMu.RUnlock()
+	return gs.inFlight != nil
+}
+func injectGatewayNudge(ctx context.Context, text string) {
+	gs := activeGatewaySession()
+	if gs == nil || !gs.IsReady() {
+		log.Printf("[status] dropping nudge: gateway session not ready")
+		return
+	}
+	if !gs.turnInFlight() {
```
AnaBerg: **Warning — TOCTOU between `turnInFlight()` and the `sessions.send` dispatch.**

```go
if !gs.turnInFlight() {
    log.Printf("[status] dropping nudge: no turn in flight")
    return
}
reqCtx, cancel := context.WithTimeout(ctx, 15*time.Second)
defer cancel()
if _, err := gs.sendReq(reqCtx, "sessions.send", ...); err != nil {
```

`turnInFlight()` takes and releases `infMu` before returning, so the turn can complete between the check and the send — the nudge then lands outside any turn, which is the state this guard exists to prevent. The window is narrow, but it's precisely the boundary the watchdog produces (nudges fire near the end of a long turn).

Either hold the per-session synchronization across check + dispatch, or capture a turn generation/token under `infMu` and revalidate it immediately before `sendReq`. The existing timeout and dropped-nudge logging should stay as-is.

pkg/hub/watchdog_enforcement_test.go
```diff
@@ -221,3 +243,329 @@ func TestWatchdogSilentDeathStopsClaw(t *testing.T) {
...
+	}
+	s.checkClawStatus()
+	eventuallyWatchdog(t, func() bool {
+		var n int
+		_ = db.QueryRow(`SELECT COUNT(*) FROM messages WHERE id='nag'`).Scan(&n)
+		return n == 0
+	}, "forced finish nag deletion")
+}
+
+func TestRestartMidTurnEnqueuesExactlyOneResume(t *testing.T) {
```
AnaBerg, line 305: **Info — missing coverage for the `restart_count` reset case.**

`TestRestartMidTurnEnqueuesExactlyOneResume` only exercises increments from a zero baseline. Add a case that sends `beat(5)` as the baseline and then `beat(0)` as the relaunch, asserting exactly one restart resume is enqueued. That's the scenario that exposes the zero-value map read flagged in `server.go` — today it would assert 0 resumes.

The existing duplicate and subsequent-increment assertions should be kept; this can be a focused sibling test reusing the same watchdog setup and count helper.

pkg/hub/server.go
AnaBerg: Fixed in b8277ad — existence-aware lookup (`lastResumedCount, resumeRecorded := ...`); a relaunch reset to 0 now enqueues the resume, covered by TestRestartCountResetMidTurnEnqueuesResume.

pkg/hub/pr_watcher.go
AnaBerg: Fixed in b8277ad — the dedup check now fails closed: Scan error logs and returns instead of falling through to injection.

cmd/claw-bridge/main.go
AnaBerg: Fixed in b8277ad — new sendTurnScopedReq holds infMu across the in-flight check AND the WS write (5s write deadline so a wedged socket can't stall readLoop writers), awaiting the response outside the lock; dead turnInFlight removed.

pkg/hub/watchdog_enforcement_test.go
AnaBerg: Added in b8277ad — TestRestartCountResetMidTurnEnqueuesResume: baseline beat(5), relaunch beat(0) mid-turn asserts exactly one resume, duplicate beat(0)s assert no extras. Verified it fails against the previous zero-value map read.

pkg/hub/pr_watcher.go
```diff
@@ -1049,6 +1049,17 @@ func (s *Server) injectMessage(clawID, content, role string) {
 	if err := s.db.QueryRow(`SELECT tenant_id FROM claws WHERE id=?`, clawID).Scan(&tenantID); err != nil {
 		return
 	}
+	var pendingDupes int
+	// Fail closed: injecting on a failed dedup check is the outcome this
+	// guard exists to prevent.
+	if err := s.db.QueryRow(`SELECT COUNT(1) FROM messages WHERE claw_id=? AND role=? AND content=? AND delivered_at IS NULL`, clawID, role, content).Scan(&pendingDupes); err != nil {
+		log.Printf("[pr-watcher] dedup check for claw %s failed: %v", shortID(clawID), err)
+		return
```
AnaBerg: **CodeRabbit (major):** the dedup check fails closed, so a transient DB error silently drops the injection — the watchdog nudge / recovery prompt is never persisted and never retried.

The fail-closed choice is deliberate (see the comment here, and `b8277ad`), but the failure probabilities are asymmetric: a transient query error is far more likely than the double-injection this guard prevents, and the cost of dropping a watchdog nudge is a stuck agent. Options:

1. Fail open — inject on query error and accept a rare duplicate.
2. Keep fail-closed but enqueue for retry so the message survives a transient error.
3. Leave as-is if the duplicate-flood risk is judged worse than a dropped nudge.

Separately: the tenant lookup just above (`pr_watcher.go:1049-1051`) has the same drop-the-message failure mode with **no log at all**, so it fails invisibly. Worth at least a `log.Printf` there regardless of which option is chosen.

AnaBerg: Fixed in 92f93bc — went with option 1 (fail open): on a dedup-query error we log and inject anyway, since a dropped nudge/resume strands the agent while the worst case of injecting blind is one duplicate pending message (the nag flood this PR fixes came from the flag re-arm bug, not from missing dedup). Also added the missing log on the tenant-lookup failure path.

PR #555, opened 2026-07-27
perf(web): fix UI lag on the agents board

AnaBerg:
Closes [AIEDEV-14](https://linear.app/nimble-la/issue/AIEDEV-14).

## Problem

With several agents streaming, the agents board became laggy. The root cause was a render cascade that re-rendered the **entire board tree at 60fps whenever any single claw was streaming**:

1. `use-typewriter.ts` `tick()` runs on `requestAnimationFrame` and published a brand-new `displayBuffers` object every frame.
2. `app/page.tsx` derived both `claws` and `mergedMessages` from `streamingBuffers`, so every card got a new `messages` identity every frame (and `mergedMessages` copied the whole message map + allocated new `Message` objects per frame).
3. Neither `ClawBoardCard` nor `SortableClawBoardCard` was memoized, and the board passed inline arrow handlers — so all cards re-rendered, not just the streaming one.

Amplifiers per card: unmemoized `MarkdownContent` (full remark/rehype parse per message per render), no windowing on the card message list, an O(n²) `isLatestActivityMessage` call inside the render map, one 1s interval per card, and a synchronous `JSON.stringify` of the message cache on every WS event.

## Changes

**Per-frame cost**
- `MarkdownContent` memoized; `components` map and remark plugins hoisted to module scope.
- `ClawBoardCard` / `SortableClawBoardCard` wrapped in `React.memo`; board handlers stabilized via `useCallback` taking `clawId`.
- Streaming text moved out of the global message map into a dedicated `StreamingMessage` child fed by `streamingBuffers[clawId]`, so the rAF tick only re-renders the streaming card. Partial text renders as plain text; the finalized WS message still renders as markdown.

**Per-render cost**
- `latestActivityMessage` computed once per render and compared by id (was O(n²)).
- Board cards windowed to the last 50 messages.
- Per-card 1s activity-age interval replaced by a single shared ticker.

**I/O jank**
- `persistMessages` debounced (1s trailing) with a flush on `pagehide`/unmount.
- `mergeClaws` reuses the previous `Claw` object (and array) when nothing changed, so the 10s poll no longer invalidates memoization.

Pure performance refactor — no intended behaviour or styling changes.

## Verification

- `npx tsc --noEmit` — clean
- `npm run build` — passes
- `npm run lint` — the 9 remaining problems in the touched files are all pre-existing (`use-hub.ts` lines 222/389/566/592 and the chat-view `scrollRef`/`` warnings), all in code this diff does not modify. Zero new lint problems.
- Implementation by gpt-5.6-sol, reviewed independently by fable-5 and gpt-5.6; 7 findings raised and fixed (incl. two build-blocking type errors, a real `onClick` runtime bug on the card back, and follow-latest auto-scroll during streaming).

Not yet verified: live visual check with multiple agents streaming against a running hub. Worth a React Profiler pass on staging to confirm only the streaming card commits per frame.

### DIFF
--- a/web/app/page.tsx
+++ b/web/app/page.tsx
@@ -6,7 +6,6 @@ import { ConversationView } from "@/components/conversation-view"
 import { SetupScreen } from "@/components/setup-screen"
 import { ManualTriggerModal } from "@/components/manual-trigger-modal"
 import { useHub } from "@/hooks/use-hub"
-import type { Message } from "@/lib/types"
 import { isConfigured, type Workflow } from "@/lib/api"
 import { requestAuthToken } from "@/lib/auth-storage"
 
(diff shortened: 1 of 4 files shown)


AnaBerg:
## Local measurements (before/after)

Reproduced the lag locally against a mock hub that streams chunks for N agents (120 messages of history each, one chunk every 60ms per agent), driving the production static export in headless Chrome. Metrics are frame timings + `PerformanceObserver('longtask')` over a 20s window.

| Scenario | FPS | frame p90 | main thread blocked | longest task |
|---|---|---|---|---|
| `main`, 6 agents | 4.6 | 250ms | 79% | 317ms |
| this branch, 6 agents | **43.2** | 33ms | **0%** | 0ms |
| `main`, 12 agents | 2.2 | 617ms | 99% | 570ms |
| this branch, 12 agents | **20.6** | 67ms | **10%** | 72ms |

Each row was run twice with consistent results (e.g. baseline 6 agents: 4.6 / 4.7 fps; branch: 43.2 / 44.9 fps).

## Regression found and fixed during verification

Visual check showed that moving the live buffer out of the message list also took it out of the markdown renderer — partial text displayed raw `**bold**`, code fences and table pipes until the message finalized. Fixed in 5c92e06: the streaming buffer renders as markdown again, but the text is sampled at 8Hz instead of every animation frame so the parse rate stays bounded. That is the cost difference between the 12-agent numbers above (21 fps / 0% blocking with plain text vs 20.6 fps / 10% blocking with live markdown) — worth it to keep the rendering correct.

PR #571, opened 2026-07-29
feat(hub): Slack notifications for agent lifecycle events

AnaBerg:
Adds the hub's first outbound notifier: Slack messages when an agent starts working, opens a PR, or fails. Every existing webhook in the hub is inbound — this is new surface.

## Design

**Dispatch is a poller over `task_run_events`, not a post-commit hook.** `recordTaskRunEventTx` is called from many places *inside* larger transactions, so hooking the higher-level `recordTaskRunEvent` would have silently missed `agent_started` and `pr_opened` — the two events that matter most. The poller has a single insertion point, survives restarts, and keeps network I/O away from the analytics transactions.

**Two scan passes, one delivery stack.** Claws attached to a task run are covered by the task-run pass; claws without one (`task_run_id = ''`, e.g. `make dev-claw`) are covered by a claw pass that derives the same three kinds from claw state and `claw_prs`. They share one client, limiter, dedupe table and thread table, with namespaced keys (`claw:<id>:...`) so the namespaces cannot collide.

Exclusivity is enforced by `task_run_id` and defended twice: a short grace period before a claw counts as ad-hoc, and a re-read of `task_run_id` immediately before each send, so a claw adopted into a run mid-flight cedes to the task-run path rather than double-notifying.

**Threading:** one Slack thread per run (or per claw), so a burst of parallel agents stays readable in the channel.

## Config

```yaml
secrets:
  slack_bot_token: xoxb-...

notifications:
  slack:
    enabled: true
    bot_token_ref: slack_bot_token
    channel: C0123ABCD
    thread_by_run: true
    poll_interval: 5s
    events: { agent_started: true, pr_opened: true, failures: true }
```

Slack side needs a bot token with `chat:write` and `chat:write.public` — no Event Subscriptions, no Socket Mode, no public URL, since the hub only ever talks *to* Slack.

## Manual trigger

`POST /api/notifications/slack/test` (admin auth) renders any event kind, from real `run_id`/`claw_id` context or synthetic sample data. `dry_run: true` returns the Block Kit payload without contacting Slack — the intended way to iterate on formatting.

## Verification

Validated end to end against a local dev hub with a real Slack workspace: dry runs for all three kinds, a real send, and a live ad-hoc claw whose `agent_started` was delivered by the poller (`claw:5dbcd607-...:agent_started | sent`).

Two review passes found and fixed, among others:
- a poller wedge once a burst exceeded the batch size — the cursor could never advance and **all** future events were dropped permanently
- `observed_at` being a provider timestamp, not an insertion time, so backdated `pr_opened` rows landed behind the cursor and were never sent
- readiness gated on `bootstrap_ok`, which only the VM provisioners set — this excluded every docker/local claw, i.e. the exact case the claw pass exists for

## Message design

Each event type carries its own emoji, a plain-English title and a colour stripe (blue started, green PR opened, red hard failures, amber soft outcomes). Blocks are wrapped in a single attachment to get the stripe; layout is headline -> subject -> blockquoted reason -> dim context metadata holding the raw type.

| Event | | Title | Colour |
|---|---|---|---|
| `agent_started` | :rocket: | Agent started | blue |
| `pr_opened` | :tada: | PR opened | green |
| `agent_stopped` | :skull: | Agent died | red |
| `creation_failed` | :no_entry_sign: | Couldn't create the agent | red |
| `provision_failed` | :construction: | Couldn't get a machine | red |
| `bootstrap_failed` | :boom: | Agent crashed during startup | red |
| `permission_or_auth_failed` | :lock: | Agent was denied access | red |
| `provider_lost` | :satellite_antenna: | Lost contact with the provider | red |
| `timeout` | :hourglass_flowing_sand: | Agent ran out of time | amber |
| `done_without_pr` | :mailbox_with_no_mail: | Agent finished without a PR | amber |
| `unknown_failure` | :question: | Agent failed | red |

Two things worth calling out. The push-notification fallback is emoji-led with consistent `where / what / why` slots and includes the failure reason — on a phone it is the whole message. And with attachments the top-level `text` is *not* a hidden fallback: Slack renders it as a visible line above the card, duplicating the headline, so the summary lives in the attachment's `fallback` field.

A test fails if any two event types share an emoji or title, or if a title leaks snake_case.

## Pluggable providers

Notifications go through `pkg/hub/notify`: a provider-agnostic `Message`, a `Notifier` interface, and a registry whose entries carry the secret-setting names the doctor validates against — so a new provider (email, Teams) is one file plus one registry entry and inherits doctor coverage. Slack is the first provider; `Severity` maps to the attachment stripe inside it, so no hex leaves the provider. Design follows the closed PR #477, with a richer `Message` so the per-event emoji/colour/hierarchy survives.

Notifiers are hub-level and named, referenced by `via`:

```yaml
notifications:
  notifiers:
    eng-agents: { type: slack, token_secret: slack_bot_token, channel: C0123ABCD }
  lifecycle:
    via: eng-agents
```

## Workflow stages can notify

```yaml
on_enter:
  notify:
    via: eng-agents
    text: "Merged {{.Issue.Identifier}}: {{.Issue.URL}}"
```

Reuses the existing `inject` templating helper, resolves through `Server.notifierFor` so it shares the cached instance and per-channel pacing with the lifecycle notifier, and is non-blocking: a failed notification never blocks a stage transition.

## Review

Reviewed by fable-5, gpt-5.6-sol and CodeRabbit independently. **Each found defects the other two missed**, and two reviewers converged on the same delivery bug from different angles. Findings were triaged against the code before being fixed — two were refuted on their own premises rather than patched. Fixes were confirmed by mutation: revert the change, watch the test go red.

Notable fixes, all with regression tests:

- muted event categories flooded the channel on re-enable, or silently dropped, depending on timing
- the ad-hoc claw PR cursor used SQLite rowids from a table whose rows are deleted on seven paths — reused rowids permanently skipped PRs
- Slack pacing had regressed from process-wide to per-instance; two sends that should be 1s apart completed in 16ms
- a proxy echoing the `Authorization` header put the bot token into logs, an HTTP response, and the claw conversation
- untrusted issue titles were interpreted as mrkdwn, so `<!channel>` in a title would have pinged a whole channel from a trusted sender
- `dry_run` and the real send rendered from different settings, so the preview could lie about the destination

## Agent idle

The fourth notification: an agent that stopped making progress. The signal is `lastTurnFinishedAt` + `!isBusyLocked()`, not `lastStatusAt` — the bridge refreshes that one with `status_pong`, so it measures whether the *process* is alive and never fires for a stalled agent.

Detecting idleness is easy; not crying wolf is the hard part. An agent that delivered a PR and is waiting on review is idle and fine, and alerting on it would get the whole feature muted. Review found the first exclusion rule too permissive in five ways — most importantly that an interactive ad-hoc claw would alert every time a human stopped typing.

`event_type` is CHECK-constrained, so this ships a rebuild-and-copy migration that preserves rowids explicitly: the notifier watermark is a rowid cursor, and an implicit copy would renumber every row and silently skip or replay notifications.

## Config errors are caught when you write them

`via` is validated on every path that persists a pipeline — workflow push, workspaces push, factories push, settings. Not at `pipeline.Parse`: a failing parse makes the pipeline unusable and the agent never runs, so a typo in a Slack message would cost an entire automated run. Runtime stays lenient and warns per send.

## Known gaps

- `HubConfig.Validate()` has no production caller, so an invalid Slack block is not rejected at load. The notifier re-validates each tick and pauses with a logged warning, and the test endpoint surfaces the error — but the feedback is in the log, not at boot.
- No settings-API/UI surface for `notifications`; config is edited in `hub.yaml`.
- No settings-API or UI surface for `notifications`; config is edited in `hub.yaml`. Out of scope by decision.
- Author-time validation checks that `via` names a defined notifier. A notifier deleted or renamed *after* a workflow is saved strands it — that drift is the doctor's job, and the doctor is on-demand.


pkg/hub/lifecycle_notifier.go
```diff
@@ -0,0 +1,1244 @@
...
+}
+
+// recordNotificationDelivery persists one delivery row. A failed write is
+// stashed in memory and retried on later ticks; while stashed, the delivery
+// key still counts as handled (see lifecycleDeliveryPending) so the message
+// cannot be re-sent externally. This matters most for claw-state events,
+// which have no rowid cursor: the delivery row is their only dedupe, so
+// logging and forgetting a failed write meant a duplicate send every tick
+// until the DB accepted the insert.
+func (s *Server) recordNotificationDelivery(eventID, runID, messageHandle, status string) {
```
AnaBerg, line 674: **[major — test gap]** The pending-delivery stash is the only guard against duplicate external sends when the delivery-row INSERT fails, and it has zero test coverage: `lifecyclePendingDeliveries`, `flushPendingNotificationDeliveries`, and `recordNotificationDelivery` have no hits in any `*_test.go`. All lifecycle tests run against a healthy DB. The infra (fake Slack server + settable DB) already exists, so a test that fails the insert and asserts (a) no re-send next tick, (b) drain on recovery, is cheap.

Related **[minor]**: the stash only closes the dedupe gap within a single process lifetime. For claw-pass events (no rowid cursor — the delivery row is their only dedupe, per the comment above), a crash after a successful Send but before the flush drains the stash re-selects the claw via the SQL anti-join on next boot and sends the Slack message again. The crash/restart window is exactly when SQLite writes tend to fail (shutdown, disk-full).

pkg/hub/agent_idle.go
```diff
@@ -0,0 +1,382 @@
...
+		return
+	}
+	// anchor identifies the stretch: the end of the last real turn, or — for
+	// a claw that never ran one — the moment this connection registered. A
+	// claw that came up connected and was never prompted at all (workflow
+	// start failed, pipeline entry stage never injected, trigger dropped the
+	// issue) is the most common real stall, so "no turn ever" must alert too.
+	anchor := lastTurn
+	if anchor.IsZero() {
+		anchor = connectedAt
```
AnaBerg: **[minor — duplicate alerts]** For a claw that never ran a turn (`lastTurnFinishedAt` zero — the "most common real stall" this code targets), the anchor falls back to `connectedAt`, which is re-stamped `time.Now()` on every WS registration (server.go:2320). After any hub restart or bridge reconnect the new anchor is far past the persisted `idle_since`, the durable-latch check at line 179 fails to match, and a brand-new stretch is latched → the same unchanged stall re-alerts ~`idle_after` after every reconnect. This contradicts the design comment that "a hub restart cannot re-notify the same stretch" — that guarantee only holds when a prior turn exists to re-seed the anchor. A flaky bridge on a permanently-stalled claw yields repeated duplicate alerts.

Aside: since `stretchStart >= anchor` always, the second disjunct at line 179 is implied by the first — the OR is redundant.

pkg/hub/lifecycle_notifier.go
```diff
@@ -0,0 +1,1244 @@
...
+	// agentIdleBaseline.
+	if _, found, err := s.notifierStateInt64(agentIdleBaselineKey); err == nil && !found {
+		s.setNotifierStateInt64(agentIdleBaselineKey, epochMillis(now()))
+	}
+}
+
+// startLifecycleNotifier launches the background loop that turns lifecycle
+// events into notifications. The goroutine idles cheaply while notifications
+// are disabled and picks the config up again on the next tick.
+func (s *Server) startLifecycleNotifier() {
```
AnaBerg, line 262: **[minor — shutdown race]** This goroutine loops forever with no stop channel/context, while `Server.run` closes the DB after HTTP drain (server.go:463). A tick in flight at shutdown can complete an external Slack Send and then fail the delivery-row insert against the closed DB; the stash entry dies with the process, so the event re-sends after restart. The other hub loops (statusWatchdog, runReaper) have shutdown plumbing — this one should too, and it would also shrink the crash-window duplication noted at line 637.

pkg/hub/notify/slack.go
```diff
@@ -0,0 +1,584 @@
...
+		minSendInterval: interval,
+	}, nil
+}
+
+// Destination names the configured channel for the caller's bookkeeping
+// (see notify.DestinationReporter).
+func (s *slackNotifier) Destination() string { return s.channel }
+
+// destinationFor resolves the channel a message will post to.
+func (s *slackNotifier) destinationFor(msg Message) string {
```
AnaBerg, line 208: **[minor — security]** `newSlack()` strictly validates the configured `channel` (`^[CGD][A-Za-z0-9]+$`), but the per-message override path doesn't: `destinationFor()` returns `msg.Target` verbatim and `RenderPayload()` writes it straight into `payload["channel"]`. The only producer of a non-empty Target is the pipeline notify action, where it is template-rendered from untrusted sources (issue titles from GitHub/Linear/Jira, `{{.Inputs.*}}` from the non-admin manual-trigger endpoint). A workflow that templates `target:` lets whoever controls tracker content redirect notifications to an arbitrary channel. Suggest applying the same channel-ID validation to the resolved Target.

pkg/hub/notify/slack.go
```diff
@@ -0,0 +1,584 @@
...
+// lock screen it is 100% of what the user sees. The emoji leads (shortcodes
+// render in Slack push notifications) so the discriminator survives
+// truncation, and every message keeps the same slot order:
+//
+//	<emoji> <title> — <summary parts joined by middle dots>
+//
+// with empty parts dropped, so a slot never silently changes meaning. iOS
+// truncates at roughly two lines, so callers must front-load what matters in
+// Message.Summary and keep parts short.
+func slackFallbackText(msg Message) string {
```
AnaBerg, line 501: **[minor — security]** `slackFallbackText()` is the one render path that skips `slackEscape()`: it concatenates `msg.Title`, `msg.Text`, and `msg.Summary` raw into the attachment `fallback`. Those values come from tracker-influenced text (issue titles, failure reasons — see buildFailureMessage / buildAgentIdleMessage). The escaping exists precisely so "a crafted issue title could [not] smuggle a `<!channel>` broadcast" — the fallback field should get the same treatment, especially since the semantic tier deletes the top-level `text`.

pkg/hub/lifecycle_notifier.go
```diff
@@ -0,0 +1,1244 @@
...
+		s.parkLifecycleWatermark()
+		s.parkLifecycleClawState()
+		return
+	}
+	if err := types.ValidateNotificationsConfig(cfg); err != nil {
+		s.logPollWarningOnce("notify-config", "[notify] invalid notifications config — notifications paused: %v", err)
+		return
+	}
+	lc := cfg.Lifecycle
+	nc := cfg.Notifiers[lc.Via] // exists: validated above
```
AnaBerg: **[minor — silent failure]** `types.ValidateNotificationsConfig` resolves the lifecycle via after `strings.TrimSpace`, but both runtime consumers look it up with the raw string (here and in `handleNotificationTest`). A hub.yaml with `via: "eng-agents "` passes validation, passes doctor's `checkNotifications` (which never resolves via), and then yields a zero-value NotifierConfig at runtime — notifications silently stop while everything reports green. Trim once at load, or look up trimmed here.

pkg/hub/notify_validate.go
```diff
@@ -0,0 +1,156 @@
...
+		return nil
+	}
+	var problems []string
+	for _, ref := range notifyActionRefs(pipelineYAML) {
+		if ref.Via == "" {
+			problems = append(problems, fmt.Sprintf(
+				"stage %q has a notify action without a \"via\" (the name of a notifier under notifications.notifiers)", ref.StageID))
+			continue
+		}
+		if _, ok := notifiers[ref.Via]; !ok {
```
AnaBerg: **[minor — incomplete validation]** The author-time rejection of an unknown notify `via` checks membership in the notifiers map but never `notify.Supported(nc.Type)` (and `ValidateNotificationsConfig` only requires a non-blank type). So `notifiers.eng: {type: teams}` + a stage with `via: eng` saves with a 200, and the failure only surfaces at runtime as a per-send warning in the claw conversation. Since this branch's stated goal is author-time rejection, the type check belongs here too.

pkg/hub/pipeline/pipeline.go
```diff
@@ -270,6 +270,36 @@ type JudgeAction struct {
...
+type NotifyAction struct {
+	Enabled bool `yaml:"-" json:"-"`
+	// Via is the name of the hub notifier to send through.
+	Via string `yaml:"via"`
+	// Text is the message body (templated). It is rendered as plain text:
+	// providers escape their own markup control characters, because Text
+	// interpolates untrusted data (issue titles, command output).
+	Text string `yaml:"text"`
+	// Subject is used by providers that have one (e.g. email); templated.
+	Subject string `yaml:"subject,omitempty"`
```
AnaBerg, line 300: **[minor — surprising behavior]** The doc says Subject is "used by providers that have one (e.g. email)", implying Slack ignores it. It doesn't: a non-empty Subject makes `Message.Semantic()` true, which flips `RenderPayload` into the Block Kit attachment branch and deletes the top-level `text` supplied via `options`. Adding one optional field to an otherwise-identical notify action silently changes the visual shape of the Slack message. Either exclude Subject from the `Semantic()` trigger or document the Slack behavior here.

pkg/hub/server.go
```diff
@@ -422,6 +490,7 @@ func (s *Server) registerRoutes(mux *http.ServeMux) {
 	mux.HandleFunc("/api/settings/github/test", s.withWebAdminAuth(s.handleGitHubAppTest))
 	mux.HandleFunc("/api/settings/model-auth/login", s.withWebAdminAuth(s.handleModelAuthLogin))
 	mux.HandleFunc("/api/settings/model-auth/login/{id}", s.withWebAdminAuth(s.handleModelAuthLoginStatus))
+	mux.HandleFunc("/api/notifications/slack/test", s.withWebAdminAuth(s.handleNotificationTest))
```
AnaBerg: **[minor — API naming]** `handleNotificationTest` is provider-agnostic (resolves `cfg.Lifecycle.Via`, renders via the PayloadRenderer interface with a generic fallback), but the route bakes in `slack`. The historical `slack_*` table names have an explicit not-worth-renaming justification; a brand-new public route doesn't. `/api/notifications/test` keeps the second provider (email/Chat/Teams per the package doc) from inheriting a misleading URL or forcing a breaking rename.

pkg/hub/notify/notify.go
```diff
@@ -0,0 +1,277 @@
...
+func ConfigError(err error) error {
+	if err == nil {
+		return nil
+	}
+	return &classedError{err: err, class: ErrorConfig}
+}
+
+// PermanentError marks err as specific to this message and unrecoverable by
+// retry (malformed or oversized payload): callers record it and move on.
+func PermanentError(err error) error {
```
AnaBerg, line 211: **[minor — dead API]** Of the three error-class wrappers, only `ConfigError` has a production caller. `PermanentError` and `TransientError` are referenced nowhere outside their definitions and their own test — the Slack provider classifies via `*slackAPIError.NotifyErrorClass`, and unclassified errors already default to transient. Suggest dropping the two unused wrappers until a second provider needs them.

pkg/hub/notify/slack.go
```diff
@@ -0,0 +1,584 @@
...
+	if len(runes) <= limit {
+		return s
+	}
+	return string(runes[:limit])
+}
+
+func truncateForLog(s string, n int) string {
+	if len(s) <= n {
+		return s
+	}
```
AnaBerg, line 602: **[minor]** `truncateForLog` slices by byte (`s[:n]`) while the rest of the file truncates by runes (`truncateRunes`, `slackClampBlockText`). The input is an arbitrary HTTP error body that rides into errors surfaced on the test endpoint and injected into claw conversations; a cut mid-rune yields invalid UTF-8 there. The correct helper is already in this file.

pkg/hub/lifecycle_notifier_test.go
```diff
@@ -0,0 +1,1058 @@
...
+	if req.ThreadTS != "" {
+		t.Fatal("pr_opened without a prior root should post top-level")
+	}
+	// With no agent_started root, the pr_opened message becomes the run's root.
+	var threadTS string
+	if err := db.QueryRow(`SELECT thread_ts FROM slack_run_threads WHERE run_id='run-1'`).Scan(&threadTS); err != nil {
+		t.Fatalf("thread root row: %v", err)
+	}
+	// The muted event is parked as skipped so a later re-enable can never
+	// replay it (see TestSlackNotifierMutedCategoryNotReplayedOnReenable).
```
AnaBerg, line 241: **[minor — test gap]** This comment cites `TestSlackNotifierMutedCategoryNotReplayedOnReenable` as proof that a parked event is never replayed, but no such test exists in the repo. The test only asserts the 'skipped' row was written; it never re-enables the category and ticks again, so the actual no-replay invariant is untested for the task-run pass (the claw-pass equivalent does have one). Either add the cited test or fix the reference.

pkg/hub/lifecycle_notifier.go
```diff
@@ -0,0 +1,1244 @@
...
+		return true, false
+	default:
+		// The streak is persisted (not an in-memory counter) because the cap
+		// exists precisely for crashloop/deploy conditions: a hub restarting
+		// more often than the cap window would reset an in-memory counter on
+		// every boot, so the poisoned event would never be burned and the
+		// cursor would stay wedged for as long as the restarts continued.
+		stateKey := lifecycleTransientFailureStateKey(deliveryKey)
+		count, _, stateErr := s.notifierStateInt64(stateKey)
+		if stateErr != nil {
```
AnaBerg, line 497: **[minor — test gap]** Two error paths in the retry machinery are untested: (1) the `stateErr != nil` branch here — the deliberate choice not to increment the streak when `slack_notifier_state` is unreadable. A future refactor that falls through to `count++` would re-arm the cap on every failed read, so the poisoned event is never burned and the cursor stays wedged — the exact crashloop the persisted streak exists to prevent — and CI would stay green. (2) A transport-level failure in `slackNotifier.Send` (vs. an HTTP error status). Both are cheap with the existing fakes.

pkg/hub/notify/slack.go
AnaBerg: Fixed in 88b8d844. \`RenderPayload\` now validates a non-empty \`Target\` against the same \`slackChannelIDRegex\` \`newSlack\` applies to the configured channel, after trimming, returning a \`PermanentError\` — message-specific, never retried, notifier stays healthy.

\`TestSlackRejectsInvalidTargetOverride\` covers \`#general\`, arbitrary tracker text, a JSON-smuggling attempt, \`@here\` and user IDs, and asserts zero HTTP calls plus \`ErrorPermanent\`. A padded but valid ID still posts, since template whitespace should not break a legitimate target.

pkg/hub/notify/slack.go
AnaBerg: Fixed in 88b8d844 — and the first cut of this fix was itself defective, which is worth recording.

\`slackFallbackText\` now escapes Title, Text and every Summary part. But the initial test only exercised Summary: reverting the Title escaping left the suite green, reverting Text left it green, reverting both left the entire repo green. The Text branch is the reachable one, since the pipeline notify action sets Text and Subject and never Title, so the security fix would have regressed silently.

The test is now three independent subtests, one hostile string per field. Verified by reverting each escape individually: each turns exactly its own subtest red.

pkg/hub/agent_idle.go
AnaBerg: Fixed in 88b8d844, and your diagnosis was right about the mechanism but the first fix I wrote was wrong.

I initially special-cased "any live latch counts as the same stretch when the claw never ran a turn". That silently broke the parked case: \`parkAgentIdleStretch\` stamps \`idle_since\` without alerting for pre-baseline stretches, and since a never-turn claw never clears the latch, such a claw went permanently mute.

The root cause is the one your comment points at — \`connectedAt\` is the wrong anchor, because it is re-stamped on every registration, so the stretch identity moves under a claw whose situation is unchanged. Never-turn claws now anchor on \`claws.created_at\`, which is stable across reconnects. The special case disappears and the ordinary comparison handles both claw kinds.

Consequence, accepted deliberately and pinned by \`TestAgentIdlePreBaselineNeverTurnStallStaysParkedAcrossReconnect\`: a never-turn stall that began before the baseline stays parked. A reconnect is not a new stall.

Redundant OR disjunct removed as you noted.

pkg/hub/lifecycle_notifier.go
AnaBerg: Fixed in 88b8d844. The stash now has coverage via a SQLite \`BEFORE INSERT RAISE(ABORT)\` trigger on \`slack_notification_deliveries\`, which fails only the delivery write while selects and sends proceed — exercised through a claw-pass event, whose delivery row is its only dedupe. Asserts no re-send on the next tick while the insert keeps failing, and a drain to \`sent\` once the trigger is dropped.

On the related note: you are right, and it is inherent to at-least-once delivery without a provider idempotency key. I did not build a distributed transaction for it. It is now documented honestly on \`flushPendingNotificationDeliveries\`, and the shutdown plumbing from your other comment shrinks the window for the graceful path.

pkg/hub/lifecycle_notifier.go
AnaBerg: Fixed in 88b8d844 — stop/done channels, timer plus select instead of \`time.Sleep\`, and \`Server.run\` waits (bounded, 10s) for an in-flight tick after the HTTP drain and before \`db.Close()\`.

One correction: the premise that \`statusWatchdog\` and \`runReaper\` already have shutdown plumbing does not hold. Both are \`for range ticker.C\` loops with no stop channel or context, and \`cronScheduler.stop()\` exists but has no production caller. So there was no in-repo pattern to copy and I used a plain stop/done pair. The race you describe is real and is fixed; those other loops have the same gap, which is out of scope here.

pkg/hub/lifecycle_notifier.go
AnaBerg: Fixed in 88b8d844. Both runtime consumers now look up \`lifecycle.via\` trimmed, matching how \`ValidateNotificationsConfig\` resolves it. \`TestSlackNotifierTrimsLifecycleVia\` uses \`"test-notifier "\` and asserts the tick delivers and the endpoint returns 200; reverting to the raw lookup makes it fail with zero messages.

pkg/hub/notify_validate.go
AnaBerg: Fixed in 88b8d844. \`validateNotifyVias\` now also rejects a \`via\` resolving to a notifier whose type fails \`notify.Supported()\`, so \`{type: teams}\` is a 400 at save time instead of a per-send warning later. Provider settings and secret resolution stay the doctor's job, keeping the author-time/doctor split intact.

pkg/hub/pipeline/pipeline.go
AnaBerg: Documented rather than excluded, in 88b8d844 — so this is the one where I did not take the first option you offered, and the reasoning is worth stating.

Excluding Subject from the \`Semantic()\` trigger would mean a \`Message{Text, Subject}\` renders plain and drops the subject entirely. That breaks the package's own contract that a provider never silently discards a populated semantic field, and it is exactly backwards for the future email provider where Subject is the primary field.

So the surprise you flagged is real but it is a documentation defect. \`pipeline.NotifyAction.Subject\` and \`notify.Message.Subject\` now both say plainly that on Slack, setting \`subject\` upgrades the message from a plain line to the Block Kit attachment layout, and to omit it for plain rendering. \`TestSlackSubjectOptsIntoRichRenderingWithoutDroppingText\` pins the decision: excluding Subject from \`Semantic()\` makes it fail with the subject dropped.

pkg/hub/server.go
AnaBerg: Renamed in 88b8d844 to \`/api/notifications/test\`. Agreed the historical \`slack_*\` table names are a different case — those are internal and pre-date the abstraction, while this is a new public route that would force a breaking rename the moment a second provider lands.

Repo-wide grep confirmed the only other reference was the test helper. Nothing in \`web/\` or the example workflow pointed at it.

pkg/hub/notify/notify.go
AnaBerg: Half applied in 88b8d844. \`TransientError\` is gone, and the \`ClassifiedError\` doc now says why it deliberately does not exist: transient is already the default for unclassified errors, so the wrapper could only ever be a no-op.

\`PermanentError\` stayed, because it gained a production caller in this same change set — the \`Target\` validation from your other comment returns it, so an invalid per-message target is classified message-permanent rather than retried.

pkg/hub/notify/slack.go
AnaBerg: Fixed in 88b8d844 — \`truncateForLog\` now uses the file's existing \`truncateRunes\`.

One detail worth recording for whoever touches the test: a body of only multi-byte runes would NOT catch this bug, because 300 happens to be rune-aligned for 2-byte runes. \`TestSlackHTTPErrorBodyTruncatesOnRuneBoundary\` prefixes an ASCII character to force every subsequent rune off an even offset; that prefix is load-bearing.

pkg/hub/lifecycle_notifier_test.go
AnaBerg: Fixed in 88b8d844 by adding the cited test rather than deleting the reference, since the invariant deserved coverage: mute \`agent_started\`, tick so the event is parked as \`skipped\`, re-enable, tick — no replay, while a genuinely new event of the re-enabled category still delivers.

Verified by removing the \`skipLifecycleMutedEvents\` parking call, which turns it red. You were right that the existing test only asserted the skipped row was written and never re-enabled.

pkg/hub/lifecycle_notifier.go
AnaBerg: Both covered in 88b8d844.

The \`stateErr != nil\` branch now has a test that fails if it falls through to \`count++\` — which is the scenario you describe, where the cap re-arms on every failed read, the poisoned event is never burned, the cursor stays wedged, and CI stays green. That was the most valuable of the test-gap comments precisely because the failure mode is invisible.

Transport-level \`Send\` failure is covered too, distinct from an HTTP error status.

PR #573, opened 2026-07-30
fix(hub): use go list -e and skip modules with errors

xavpaice:
The previous resilience fix only helped when go list exited non-zero but still emitted usable JSON. With placeholder versions like k8s.io/cri-streaming@v0.0.0, go list can fail before producing any JSON, so the fallback never ran.

Use go list -e so broken modules are emitted as JSON objects with an Error field instead of aborting the whole listing, and skip any module that carries an Error field.

Tests updated and passing.


xavpaice:
Incorporated into #576 along with the stale go.mod recovery fix. Closing this one to reduce review burden.

PR #575, opened 2026-07-30
[AIEDEV-15] Add scheduled agent routines

nicoprofe:
## Summary

- add Hub-managed cron routines with editable agent tasks, timezones, timeouts, overlap handling, manual runs, and run history
- add AI-assisted routine drafting using configured API-key, Ollama, or Codex auth-profile models
- add automatic readiness checks before enabling or running a routine
- identify routine agents in the dashboard and preserve completed routine logs after their sandbox terminates
- add routine APIs, UI, documentation, examples, migrations, and test coverage

## Why

ElasticClaw workflows can already be triggered manually or by external systems, but users need a first-class way to schedule repeatable agent work. Routines provide a Hub-owned schedule while keeping every execution isolated and observable.

Only `routine`-tagged successful agents use retained `completed` history. Existing non-routine workflow lifecycle behavior is unchanged. Codex-ready image and bundled bridge changes are intentionally excluded from this PR.

## User impact

Users can create a routine from a natural-language description or manually configure its task and schedule, keep it paused until ready, run it immediately, inspect prior executions, and reopen completed routine logs without keeping the sandbox alive.

## Validation

- `make build`
- `make test`
- `go build ./...`
- `npm run lint -- --quiet`
- `npm run build`
- `git diff --check`

Linear: [AIEDEV-15](https://linear.app/nimble-la/issue/AIEDEV-15)

### DIFF
--- a/pkg/hub/terminal_transitions.go
+++ b/pkg/hub/terminal_transitions.go
@@ -2,10 +2,28 @@ package hub
 
 import (
 	"database/sql"
+	"encoding/json"
 	"log"
 	"strings"
 	"time"
 )
+
+func (s *Server) successfulClawTerminalStatus(clawID string) string {
+	var tagsJSON string
+	if err := s.db.QueryRow(`SELECT COALESCE(tags, '[]') FROM claws WHERE id=?`, clawID).Scan(&tagsJSON); err != nil {
+		return "deleted"
+	}
+	var tags []string
+	if json.Unmarshal([]byte(tagsJSON), &tags) != nil {
+		return "deleted"
+	}
+	for _, tag := range tags {
+		if tag == "routine" {
+			return "completed"
+		}
+	}
+	return "deleted"
+}
 
 func (s *Server) execStatusLogged(opCtx, query string, args ...any) (sql.Result, error) {
 	res, err := s.db.Exec(query, args...)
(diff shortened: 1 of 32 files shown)


xavpaice:
I haven't read all the details of this change, but I wanted to ask a question - how is this different from the existing cron workflow trigger?  We keep logs from those (recent changes have made that easier to read), but this may add a bunch more rigor which would be helpful.

Here's an example of one I added just recently: https://github.com/replicatedhq/troubleshoot/blob/main/.factory-workflows/dependency-update.yaml

marccampbell:
@xavpaice you should be aware of this incase there's overlap with your dep work?

AnaBerg:
@xavpaice this is a feature been asked by FASTer people, we are improving the UX and observability

pkg/hub/workspace_api.go
```diff
@@ -290,6 +301,19 @@ func (s *Server) handleWorkspaceWorkflowPatch(w http.ResponseWriter, r *http.Req
 		return
 	}
 	if req.Enabled != nil {
+		if *req.Enabled && (workflow.Enabled == nil || !*workflow.Enabled) {
```
AnaBerg, line 304: This preflight guard runs on every `enabled=true` PATCH, not just routines. `preflightRoutine` fails with "Routine schedule missing" whenever `workflow.Trigger.Cron == nil` (see routine_preflight.go), so this now returns 409 when enabling a non-cron workflow (e.g. a Linear/GitHub-triggered one) that goes through the same `handleWorkspaceWorkflowPatch` endpoint. Suggest gating this block on `workflow.Trigger != nil && workflow.Trigger.Cron != nil` so only actual routines hit the preflight check. Didn't see a test covering enabling a non-cron workflow through this path.

pkg/hub/terminal_transitions.go
```diff
@@ -2,11 +2,29 @@ package hub
...
+func (s *Server) successfulClawTerminalStatus(clawID string) string {
+	var tagsJSON string
+	if err := s.db.QueryRow(`SELECT COALESCE(tags, '[]') FROM claws WHERE id=?`, clawID).Scan(&tagsJSON); err != nil {
+		return "deleted"
+	}
+	var tags []string
+	if json.Unmarshal([]byte(tagsJSON), &tags) != nil {
+		return "deleted"
+	}
+	for _, tag := range tags {
```
AnaBerg, line 20: Routine identity is derived from the `"routine"` tag at terminal-transition time, but tags are freely editable by users via the TagEditor at any point during the run. Removing the tag mid-run flips the terminal status to `deleted` (history disappears from the UI); adding it to an unrelated claw makes it terminate as `completed` instead. Worth considering a dedicated non-editable field (e.g. an `origin`/`kind` column set at creation, matching `triggerKind` in workflow_creator.go) rather than relying on a mutable tag for this.

web/components/routines-section.tsx
```diff
@@ -0,0 +1,604 @@
...
+              <Badge className="gap-1 border-emerald-500/30 bg-emerald-500/10 text-[10px] text-emerald-700 dark:text-emerald-300">
+                <CheckCircle2 className="size-3" />
+                ready
+              </Badge>
+            ) : preflight ? (
+              <Badge className="gap-1 border-amber-500/30 bg-amber-500/10 text-[10px] text-amber-700 dark:text-amber-300">
+                <AlertTriangle className="size-3" />
+                needs setup
+              </Badge>
+            ) : null}
```
AnaBerg, line 299: When `fetchRoutinePreflight` rejects (e.g. network error), `preflight` stays `null` and `checking` is `false`, so this ternary chain falls through to `null` — no badge, no error message. The Enable toggle and "Run now" button are still disabled via `!preflight?.ready` (lines ~338/349), so the user sees a routine that's silently stuck with no indication of why. Consider distinguishing "preflight failed to load" from "preflight ran and found issues" and surfacing the former as an error state.

web/components/routines-section.tsx
```diff
@@ -0,0 +1,604 @@
...
+      setRefreshVersion((value) => value + 1)
+    } catch (err) {
+      setError(err instanceof Error ? err.message : "Failed to start routine")
+    } finally {
+      setBusyRoutine("")
+    }
+  }, [])
+
+  const addRoutine = useCallback(async (draft: RoutineDraft) => {
+    if (workflows.some((workflow) => workflow.name.toLowerCase() === draft.name.toLowerCase())) {
```
AnaBerg, line 121: Duplicate-name check is client-side only, against a `workflows` list that may be stale (no re-fetch immediately before this check), and case-insensitive. Since the underlying POST appears to upsert by name, a stale list could let two routines race to the same name and one silently overwrite the other server-side. Might be worth enforcing uniqueness server-side (409 on conflict) rather than relying on this client guard.

PR #580, opened 2026-08-03
feat(web): promote runs and analytics to top-level routes and gate cost behind a permission

AnaBerg:
Phase 1 of the UX restructuring: fix the information architecture so each persona has a reachable home, and introduce the permission seam that makes cost visibility a policy rather than a hidden screen.

The cost gate and moving analytics out of Settings ship together on purpose. Task 2 is what lets non-admins reach analytics at all; shipping it without the gate would open a window where any authenticated user sees spend.

## What changed, per persona

**Developer** — gains run history that was previously unreachable. `/analytics` is now open to any authenticated user, and its Delivery tab carries the runs table plus a detail panel with attempts, events, PRs and agent logs — so a terminated agent no longer just disappears from the board. Sees no cost or token figure anywhere.

**Tech lead** — `/analytics` is now a real top-level route instead of a 28-line redirect stub into the admin-gated Settings SPA. Delivery and Cost are split into two tabs, so delivery reporting can be shared without also handing over spend.

**Admin** — unchanged reach, plus lands on `/analytics` after sign-in. `/settings` now has an actual route guard: previously a non-admin who typed the URL got the full settings chrome, nav and workspace picker, then a bare "Failed to load" when `/api/settings` returned 403.

## The `costs:read` capability

Enforcement is server-side. The frontend only reads the capability to decide what to render — it is not the boundary.

- `/api/analytics/costs` and `/api/analytics/cost-drivers` → `403`
- `/api/analytics/runs` and `/runs/{id}` → `estimatedCostUsd`, `inputTokens`, `outputTokens`, `totalTokens`, `model` are **omitted**, not zeroed. A zero cost is a lie; an absent field is the truth.
- `/api/analytics/effectiveness` → `costPerMergedPr` and `topTicketsByCost` omitted, every delivery field kept
- `/api/analytics/filter-options` → the `models` inventory is omitted, and a `model=` filter is rejected with 403. Without this, the model attribution redacted from the runs payload was trivially recoverable by filtering on a sibling endpoint.

`/api/auth/me` now returns `capabilities` alongside `is_admin`, which is kept for backward compatibility:

```json
{"auth_method":"github","login":"carol","is_admin":false,"capabilities":{"costs:read":true}}
```

### New config key

`auth.access.cost_viewers` is optional. Absent means "admins only".

```yaml
auth:
  access:
    admins:
      - alice
    # Optional: grant costs:read without granting admin.
    cost_viewers:
      - carol
```

## Phase 1 does not introduce users, roles or teams

This is deliberate and worth stating plainly. `costs:read` currently resolves to "admin, or listed in `cost_viewers`", and **password sessions remain full admin** — they resolve to `costs:read: true`, unchanged from today. Phase 2 replaces the resolver with a real role model. The seam is intentionally narrow so that only `canViewCosts` changes.

## Verification

`go build ./...`, `go test ./pkg/hub/...`, `npm run lint` (0 errors; 2 pre-existing `` warnings in untouched files) and `npm run build` all pass. The static export produces `/analytics` and `/runs` as prerendered routes.

Both states were then exercised against a live hub seeded with 48 runs, driving the real UI in a browser.

**With `costs:read`** (admin `alice`, and non-admin `carol` via `cost_viewers`, and password sessions) — `/analytics` shows both Delivery and Cost tabs plus the Model filter, the runs table shows Model and Cost columns, and the run detail panel shows the Usage block (prompt/completion/total tokens, estimated cost, model).

**Without it** (GitHub session `bob`, in neither list) — no Cost tab, no Model filter, no cost columns, no Usage block, and the inline notice "Cost analytics requires the `costs:read` permission." in the Cost tab's place. The app neither logs out nor shows an error banner. The network panel confirms no cost request is issued at all: only `summary`, `effectiveness`, `general-stats`, `runs` and `filter-options` are fetched.

Response bodies were scanned directly rather than trusting the UI. For a caller without the capability, every one of `/api/auth/me`, `/analytics/effectiveness`, `/analytics/runs`, `/analytics/runs/{id}`, its `attempts`/`events`/`prs`/`outputs` subresources, `/analytics/summary`, `/analytics/general-stats`, `/analytics/filter-options` and `/analytics/factories` contains no `estimatedCostUsd`, `inputTokens`, `outputTokens`, `totalTokens`, `usageUpdatedAt`, `model`, `models`, `costUsd`, `costPerMergedPr` or `topTicketsByCost` key. `costs` and `cost-drivers` return `403 {"error":"costs:read permission required"}`, and `?model=` is rejected with 403 on the four ungated endpoints. The same keys are present for a capable caller.

The legacy URL `/settings/{ws}/workspace-analytics?workflow=review&status=failed` redirects to `/analytics?workflow=review&status=failed&workspace=eng` with filters applied, for admins and non-admins alike.

Screenshots of `/analytics` and the agents board in both states are attached separately — GitHub image uploads need an authenticated browser upload that the automated run could not perform.

### Navigation

A later commit consolidates navigation into a single full-width `AppHeader` above the sidebar and content, carrying Agents / Analytics and the admin-only Settings gear. It replaces three separate nav surfaces (a collapsed sidebar rail, an expanded sidebar variant, and a bespoke top bar on `/analytics`) that made the nav appear to move between screens. The same commit removes the standalone `/runs` screen as redundant with the analytics runs table.

Verified against the **production static export** served with the hub's SPA-fallback and API-proxy behaviour, not just the dev server: `/analytics` renders with both tabs for a capable session and Delivery-only plus the policy notice for a non-capable one, the board renders with the header above the sidebar, the gear appears only for admins, and client-side navigation between Agents and Analytics keeps the active highlight correct in both directions.

## Non-goals respected

No schema change or migration, no user/role/team model, no API token product, no workflow authoring UI, and no new npm or Go dependency (`go.mod`, `package.json` and `package-lock.json` are untouched).


AnaBerg:
### CI note — the three E2E failures were runner contention, not a code defect

`E2E / Daytona + Linear`, `E2E / Replicated + GitHub Issues` and `E2E / exe.dev + GitHub Issues` all failed with the same signature:

```
e2e_test.go:241: timed out waiting for agent <uuid> to reply
```

Each ran for ~10 minutes and hit the test's own timeout. What isolates the cause is how they behave alone versus together:

| run | concurrency | result |
| --- | --- | --- |
| original | 8 E2E jobs at once | 5 pass, these 3 time out at ~10m |
| retry #1 | all 3 retried together | all 3 time out again at ~10m |
| retry #2 | **Daytona + Linear alone** | **pass, 5m16s** |
| retry #3 | **exe.dev + GitHub Issues alone** | **pass, 2m22s** |
| retry #4 | **Replicated + GitHub Issues alone** | **pass** |

Run concurrently they time out; run one at a time the exact same commit passes in a fraction of the budget. That points at contention for a shared resource these jobs need in order to get an agent reply — not at determinism in the code.

Supporting evidence that the branch is not at fault:

- Five other agent-driven E2E variants passed on this same commit.
- The PR's entire non-frontend diff is `pkg/hub/server.go` (the analytics cost gate plus a static-file path mapping for the settings UI), the analytics API files and `pkg/types/template.go`. Nothing in the agent, bridge, webhook, provider or bootstrap path is touched.
- `main` is green on E2E across its recent commits (0 failures out of 11), so the suites themselves are healthy.

**All 15 checks are now green.**

Worth flagging for the repo generally: the agent-driven E2E variants do not appear safe to run at this concurrency, and will keep producing false reds on unrelated PRs until that is bounded.

PR #590, opened 2026-08-05
feat(workspace): add validate command to check workspace definitions without pushing

xavpaice:
Adds `elasticclaw workspace validate [name]` to validate workspace definitions locally without pushing them to the hub.

- Uses the same collection and validation logic as `workspace push`, but stops after validation and prints a summary.
- Supports the same filtering and --path options as `workspace push` for consistency.
- Includes tests for default path scanning, name filtering, custom path validation, and invalid workspace detection.

All tests pass: `go test ./cmd -count=1`

### DIFF
--- a/cmd/workspace.go
+++ b/cmd/workspace.go
@@ -25,6 +25,7 @@ func WorkspaceCmd() *cobra.Command {
 	}
 	cmd.AddCommand(workspaceCreateCmd())
 	cmd.AddCommand(workspacePushCmd())
+	cmd.AddCommand(workspaceValidateCmd())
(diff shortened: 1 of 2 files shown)


xavpaice:
Addressed the Greptile review: added `Args: cobra.MaximumNArgs(1)` to `workspace validate` so it rejects multiple workspace names, and added a test covering that case. Update pushed.

xavpaice:
Updated: renamed `collectWorkspacesForPush` to `collectAndValidateWorkspaces` and updated its doc comment, since the function is shared by `workspace push` and `workspace validate`. Tests pass with `go test ./cmd -count=1`.

PR #603, opened 2026-08-06
fix(hub): hide plan_gate system markers from chat

marccampbell:
## Summary
- Hide `__PLAN_GATE_ACCEPTED__:<stage>` system markers from message list + timeline APIs (they leaked into UI and copy-transcript, e.g. AMA-111)
- Route `/api/messages` through the shared `hiddenSystemMessages*` helper so filters stay consistent
- Example plan-gate workflow implement inject leads with **Starting implementation** for clearer stage UX

## Test plan
- [x] `go test ./pkg/hub/ -run TestHandleMessagesFiltersWakeMarkers`
- [ ] Run a plan_gate workflow; confirm system marker does not appear in chat or copied transcript
- [ ] Confirm wake / freeform plan markers still hidden
- [ ] Confirm implement inject still queues the agent turn

### DIFF
--- a/examples/workflows/linear-issue-plan-gate.yaml
+++ b/examples/workflows/linear-issue-plan-gate.yaml
@@ -107,7 +107,10 @@ stages:
           verdict: pass
     on_enter:
       inject: |
-        Plan accepted (status {{ .Outputs.plan.status }}).
+        ## Starting implementation for {{.Issue.Identifier}}
+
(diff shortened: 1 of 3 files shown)


marccampbell:
Addressed Greptile P2: escape `_`/`%` in the plan_gate marker LIKE prefix via `ESCAPE '\'` so underscores are literal (see `sqlLikeLiteralPrefix`).

PR #606, opened 2026-08-08
fix(cli): show remote CalVer on hub upgrade

marccampbell:
## Summary

`elasticclaw hub upgrade` printed:

```text
Remote: unknown  →  Target: 2026.8.8-beta.2
```

even when the remote hub was running a real build.

**Cause:** `parseVersionFromOutput` required a leading `v` (`elasticclaw v…`), but `elasticclaw version` prints CalVer without it:

```text
elasticclaw 2026.8.8-beta.2 (commit: …, built: …)
```

**Fix:** Parse CalVer and legacy `v`-prefixed lines (and JSON). Also probe `/usr/local/bin/elasticclaw` and `sudo` when checking the remote version.

## Test plan

- [x] `go test ./cmd/ -run TestParseVersionFromOutput`
- [ ] `elasticclaw-beta hub upgrade --profile …` shows the real remote version before upgrade

### DIFF
--- a/cmd/hub_upgrade.go
+++ b/cmd/hub_upgrade.go
@@ -76,9 +76,14 @@ func runHubUpgrade(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
-	// Check current version on server
+	// Check current version on server. Prefer PATH then the install default;
+	// sudo when needed (matches the post-upgrade version probe).
 	fmt.Println("Checking remote version...")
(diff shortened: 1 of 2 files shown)


marccampbell:
Addressed Greptile **P1**: upgrade equality now uses `sameReleaseVersion` (strip optional leading `v`) so remote `0.1.0` matches GitHub target `v0.1.0` without re-downloading. Download URL still uses the exact release tag.

PR #614, opened 2026-08-10
fix: clarify GitHub auth helper for agents

xavpaice:
Agents were reporting that 'gh' is not authenticated even though the ElasticClaw credential helper and wrapper are in place.

The root cause is that 'gh auth status' checks ~/.config/gh/hosts.yml, not the GH_TOKEN env var that our wrapper refreshes on every invocation. Because the installation token is short-lived and minted per call, nothing is persisted in hosts.yml, so gh auth status can misleadingly report "not authenticated".

This change updates the startup instructions that every agent reads:

- Explain that /usr/local/bin/gh is a wrapper that calls the credential helper before delegating to the real gh binary.
- Tell agents not to use gh auth status to verify authentication.
- Direct agents to use real commands like gh api rate_limit or gh repo view <owner>/<repo> instead.
- Repeat the existing rule: do not run gh auth login or set GH_TOKEN/GITHUB_TOKEN.

Files updated:
- .elasticclaw/workspaces/elasticclaw/AGENTS.md
- pkg/hub/server.go (injected TOOLS.md GitHub Access section)
- pkg/hub/github_webhook.go (PR-assignment BOOTSTRAP.md context)

Verification:
- go build ./...
- go test ./... -count:1


xavpaice:
Added to workspace instructions, seems to be working without this change.

PR #618, opened 2026-08-10
feat: v2 workflow history and logs

xavpaice:
Implements v2 workflow history and logs support so the UI and CLI can list v2 runs and read agent logs on par with v1, and lays groundwork for future v2 cron scheduling.

### Backend
- Adds `trigger_type` to `workflow_v2_runs`.
- Adds v2 endpoints:
  - `GET /api/v2/workspaces/{ws}/workflows/{wf}/runs`
  - `GET /api/v2/workflow-runs/{runId}/logs`
  - `GET /api/v2/workflow-runs/{runId}/attempts`
  - `GET /api/v2/workflow-runs/{runId}/attempts/{attemptId}/logs`
- History is returned as one row per attempt because each attempt has its own `claw_id`.
- v1 endpoints are unchanged.

### CLI
- `workflow runs` and `workflow logs` fetch the workflow view first and branch on `schemaVersion`.
- Added `--attempt` flag to `workflow logs` for v2 attempt-specific logs.

### Web UI
- `WorkflowRunsDialog` branches on `schemaVersion` and renders v2 history rows per attempt.
- `WorkflowRunLogsDialog` supports both v1 and v2 log targets.
- `ClawActivityLog` accepts an optional fetcher so it can render v2 logs without a claw_id.

### Docs
- ADR: `docs/adr/0001-v2-workflow-history-and-logs.md`

### Verification
- `go test ./...` passes.
- `npm run typecheck`, `npm run lint`, and `npm run build` pass in `web/`.


xavpaice:
Included in #681

PR #629, opened 2026-08-13
fix(hub): render {{.Outputs.*}} templates in cron and manual workflow injects

xavpaice:
The hub only rendered `inject` templates when a claw had an associated Linear/GitHub issue or manual inputs. Cron-triggered workflows have no issue ID, so the renderer was skipped entirely and agents saw literal `{{ .Outputs.branch_prep.status }}` text instead of the rendered values.

This change always renders inject templates with persisted pipeline outputs (and still includes Issue/Inputs when applicable). It also adds a test verifying that a cron claw can reference a prior stage's JSON output via `{{ .Outputs.*}}`.

Fixes the behavior observed in cron workflows like `ec2-update-go` where `{{ .Outputs.branch_prep.status }}` was emitted verbatim.

### DIFF
--- a/pkg/hub/pipeline_runner.go
+++ b/pkg/hub/pipeline_runner.go
@@ -1198,110 +1198,74 @@ func (s *Server) runOnEnter(clawID string, stage pipeline.Stage, ctx pipelineCon
 		injectMsg := stage.OnEnter.Inject
 		manualInputs := s.loadManualTriggerInputs(clawID)
 
-		// Render {{.Issue.Identifier}}, {{.Issue.Title}}, {{.Issue.URL}} if this is a Linear claw
-		// GitHub Issues IDs are owner/repo/number format (contain "/"), Shortcut IDs start with "sc-"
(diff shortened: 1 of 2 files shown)


xavpaice:
The `{{ .Issue.* }}` namespace is intentionally shaped for Linear/GitHub issue details (`linearIssueDetails` / `githubIssueDetails`). There is no `shortcutIssueDetails` struct, and Shortcut stories already get their context via a `CONTEXT.md` template file built by `buildShortcutContext()`.

So the old code excluded `sc-...` IDs from `{{ .Issue.* }}`. The bug was that the old code used that exclusion as a reason to skip **all** template rendering, including `{{ .Outputs.* }}`. This fix preserves the `{{ .Issue.* }}` exclusion for Shortcut while still rendering `{{ .Outputs.* }}` and `{{ .Inputs.* }}` for Shortcut, cron, and manual workflows.

marccampbell (approved):

PR #636, opened 2026-08-18
feat(AIEDEV-23): add comment_issue on_enter action for v1 pipelines

cabrera-dario-nimble:
## Summary                                                                       
                                      
  - Adds `comment_issue` as a new `on_enter` action for v1 pipelines, letting        
  workflows post a templated comment back to the originating tracker (Jira, GitHub,
  Linear, Shortcut) when a stage is entered.                                         
  - Supports scalar shorthand (`comment_issue: "body"`) and full mapping (`body`,  
  `issue_id`, `continue_on_error`). Coexists with `move_issue` in the same           
  `on_enter`; move runs before comment.                                            
  - Uses the same template context as `inject` (`{{.Issue.*}}`, `{{.PullRequest.*}}`,
   `{{.Inputs.*}}`, `{{.Outputs.*}}`).                                               
  - Hardens `queryJiraIssues`: relative JQL dates (`-Nm`) instead of ISO timestamps,
  and requests `fields: "*all"`.                                                     
  - v1 → v2 converter emits a warning for `comment_issue` and does not auto-translate
   it.                                                                               
                                                                                   
  Linear: [AIEDEV-23](https://linear.app/nimble-la/issue/AIEDEV-23)                  
                                                                                   
  ## Test plan                        
                                                                                   
  - [ ] `make test` passes (`pkg/hub/pipeline/pipeline_test.go`,                     
  `pkg/hub/pipeline_runner_test.go`, `pkg/types/convert/convert_test.go`).
  - [ ] Manual: v1 workflow with `on_enter.comment_issue: "hello"` posts a comment on
   a Jira issue in a test project.                                                 
  - [ ] Manual: same workflow against a GitHub issue, a Linear issue, and a Shortcut
  story.                                                                             
  - [ ] Manual: `continue_on_error: true` suppresses tracker failure and lets the
  stage proceed.                                                                     
  - [ ] Manual: `continue_on_error: false` on a non-terminal stage fails the stage on
   tracker error.                                                                    
  - [ ] Manual: terminal-stage tracker error is swallowed.                         
  - [ ] Manual: `move_issue` + `comment_issue` in the same `on_enter` — move runs    
  first.                                                                           
  - [ ] Manual: Jira polling (`queryJiraIssues`) still returns expected issues after
  the JQL / fields change against a live project.                                    
   
  ## Notes                                                                           
                                                                                   
  - No changes outside `pkg/`. `CLAUDE.md` and                                       
  `.elasticclaw/workspaces/jira-comment-test/` in the working tree are intentionally
  not part of this PR.                                                               
  - Additive change: existing `on_enter` YAML is unaffected because `comment_issue`
  is a new key.


pkg/hub/pipeline_runner.go
```diff
@@ -1423,70 +1555,145 @@ issueResolved:
...
+		if linearToken != "" {
+			if d, err := s.fetchLinearIssueDetails(linearToken, resolvedIssueID); err == nil && d != nil {
+				details = d
+			}
+		}
+		data["Issue"] = details
+	}
+	renderedBody := renderInjectWithData(clawID, stage.OnEnter.CommentIssue.Body, s.injectTemplateData(clawID, data))
+
+	handleErr := func(integration string, err error) error {
```
AnaBerg, line 1829: **Blocker: the default `continue_on_error: false` terminates the whole agent run.**

When this returns a non-nil error, `transitionResolvedPipelineStageWithContext` calls `stopAgentWithReason(clawID, "pipeline stage %q on_enter failed", ...)` for any non-terminal stage. So a transient 500 from the tracker, after the 3 retry attempts are exhausted, kills the claw.

Every other tracker-touching `on_enter` action — `move_issue`, `add_labels`, `remove_labels`, `close_issue` — only logs on failure and lets the stage proceed. Failing an entire workflow run because a *comment* could not be posted is disproportionate, and it is the behaviour authors get by default without opting in.

Suggestion: default `ContinueOnError` to true (i.e. invert the flag to `fail_on_error`), or at minimum keep the field but make the non-terminal path log-and-continue like its siblings.

pkg/hub/pipeline_runner.go
```diff
@@ -1423,70 +1555,145 @@ issueResolved:
...
-			return injectDelivered, nil
+			log.Printf("[pipeline] %s: no Jira tracker for connection %q, skipping comment_issue", ctx.Name(), ctx.TrackerName())
+			return nil
 		}
-		if err := s.moveJiraIssue(tracker, resolvedIssueID, targetStatus); err != nil {
-			log.Printf("[pipeline] failed to move Jira issue %s to %q: %v", resolvedIssueID, targetStatus, err)
-		} else {
-			log.Printf("[pipeline] moved Jira issue %s to %q", resolvedIssueID, targetStatus)
+		s.publishHubNotice(clawID, fmt.Sprintf("[hub] ▶ Posting comment_issue for stage %q", stage.ID))
+		if err := s.retryTrackerMove("comment jira issue", func() error {
```
AnaBerg: **Retrying a non-idempotent POST can post duplicate comments.**

`retryTrackerMove` was written for *moves*, which are idempotent — re-issuing a transition that already landed is harmless. Comment creation is not. `isPermanentTrackerMoveErr` only short-circuits 4xx, so on a 502/timeout that arrives *after* the tracker already persisted the comment, all three attempts run and the ticket ends up with 2–3 identical comments.

Same applies to the shortcut/github/linear call sites below. Options: don't retry comment creation at all, retry only on transport errors where no response was received, or dedupe by scanning recent comments before posting.

pkg/hub/jira.go
```diff
@@ -734,7 +740,7 @@ func (s *Server) queryJiraIssues(tracker workspaceIssueTracker, since time.Time,
 		requestBody := map[string]any{
 			"jql":        jql,
 			"maxResults": 100,
-			"fields":     []string{"summary", "description", "status", "labels", "assignee", "project", "updated"},
+			"fields":     []string{"*all"},
```
AnaBerg: **Blocker: `"*all"` is a payload regression and is unrelated to this feature.**

`jiraIssue` (jira.go:72) only ever decodes `summary`, `description`, `labels`, `updated`, `status`, `project`, `assignee` — exactly the list this replaces. `*all` additionally pulls every custom field, comments, attachments and so on, for up to 100 issues, on every poll tick, and all of it is discarded by the decoder.

The relative-JQL change above (`-Nm`) is a real fix and worth keeping — relative offsets sidestep the profile-timezone problem that naive datetimes have. But the `fields` change has no stated motivation tied to `comment_issue`, no test, and removes an explicit field contract. I'd revert it here and, if there is a concrete field that polling is missing, add that one field in its own PR.

pkg/hub/pipeline_runner.go
```diff
@@ -1423,70 +1555,145 @@ issueResolved:
...
+						base = "https://api.github.com"
+					}
+					if d, err := s.fetchGitHubIssueDetailsWithRetry(clawID, ghToken, repo, issueNum, base); err == nil && d != nil {
+						details = d
+					}
+				}
+			}
+		}
+		data["Issue"] = details
+	case isShortcut, isJira:
```
AnaBerg, line 1812: **`{{.Issue.Title}}` / `{{.Issue.URL}}` silently render empty for Jira.**

`fetchJiraIssue` already exists (jira.go:687) and would populate these, but this branch stubs the whole issue out to an identifier-only value. The PR description and the `CommentIssueAction` doc comment both promise "the same template data as OnEnter.Inject", which is not true for Jira — an author who copies a working `inject` body into `comment_issue` gets a comment with blank fields and no warning anywhere.

Either call `fetchJiraIssue` for the Jira case, or document that only `{{.Issue.Identifier}}` is available for Jira/Shortcut.

Separately: representing a Jira issue with a `linearIssueDetails` value is confusing to read. A small shared struct (or an interface with `Identifier`/`Title`/`URL`) would make the intent obvious.

pkg/hub/pipeline_runner.go
```diff
@@ -1324,94 +1326,224 @@ func (s *Server) runOnEnter(clawID string, stage pipeline.Stage, ctx pipelineCon
...
-				tmpl, err := template.New("issue_id").Parse(resolvedIssueID)
-				if err == nil {
-					var buf bytes.Buffer
-					data := s.injectTemplateData(clawID, map[string]interface{}{
-						"Inputs": inputs,
-					})
-					if err := tmpl.Execute(&buf, data); err == nil {
-						resolvedIssueID = buf.String()
-					}
+		goto commentIssue
```
AnaBerg, line 1551: **The eight `goto commentIssue` jumps exist only because `move_issue` was left inline.**

The new `runCommentIssueOnStage` extraction is the right shape — applying it symmetrically to `move_issue` (a `runMoveIssueOnStage` helper) lets every one of these `goto`s become a plain `return nil`, and removes the artificial `{ ... }` block that was added just to keep the jumps legal across the `resolvedIssueID` declaration. `runOnEnter` would then end with two straight-line calls instead of a label at the bottom.

pkg/hub/pipeline_runner.go
```diff
@@ -1324,94 +1326,224 @@ func (s *Server) runOnEnter(clawID string, stage pipeline.Stage, ctx pipelineCon
...
+
+	resolvedIssueID, _ := s.resolveIssueID(clawID, ctx, stage.OnEnter.CommentIssue.IssueID, issueID)
 	if resolvedIssueID == "" {
-		return injectDelivered, nil
+		return nil
 	}
 
-	// Determine issue tracker: explicit workflow/factory integration takes precedence,
-	// fall back to ID-format heuristics only when integration is empty.
+	// Determine issue tracker with the same heuristic as move_issue.
```
AnaBerg, line 1766: **Tracker-kind heuristic is now duplicated verbatim.**

This switch (integration name, falling back to `sc-` prefix / `/` presence) is byte-for-byte the same as the one in the `move_issue` path above. Two copies will drift the first time a tracker is added. Worth extracting a `resolveTrackerKind(ctx, issueID)` returning an enum, which would also make the four-way dispatch below a single `switch`.

pkg/hub/pipeline_runner.go
```diff
@@ -1324,94 +1326,224 @@ func (s *Server) runOnEnter(clawID string, stage pipeline.Stage, ctx pipelineCon
...
+	}
+	return &pipelinePullRequest{URL: prURL}
+}
+
+// resolveIssueID resolves an optional explicit issue ID (with template
+// support) against the fallback trigger issue. When explicit is empty, the
+// fallback is returned unchanged. Template rendering follows the same
+// semantics as the on_enter.inject path — {{.Inputs.*}} for manual triggers
+// and {{.Issue.*}} for automatic triggers.
+func (s *Server) resolveIssueID(clawID string, ctx pipelineContext, explicit, fallback string) (string, error) {
```
AnaBerg: `resolveIssueID` returns `(string, error)` but never returns a non-nil error — every template failure is swallowed and falls back to the unrendered value. Both call sites discard it with `_`. Either drop the error from the signature, or actually return the template parse/execute failures so callers can surface them (a silently unrendered `{{.Inputs.issue_id}}` would otherwise be sent to the tracker as a literal).

pkg/hub/pipeline/pipeline.go
```diff
@@ -381,20 +400,37 @@ func (oe *OnEnter) UnmarshalYAML(value *yaml.Node) error {
...
+	switch raw.CommentIssueRaw.Kind {
+	case 0:
+		// comment_issue not present.
+	case yaml.ScalarNode:
+		// Bare string: treat as body, no explicit issue_id or continue_on_error.
+		oe.CommentIssue = CommentIssueAction{Body: raw.CommentIssueRaw.Value}
+		oe.commentIssueRawPresent = true
+	case yaml.MappingNode:
+		if len(raw.CommentIssueRaw.Content) == 0 {
+			// Empty mapping — treat as absent so `comment_issue: {}` is a no-op.
```
AnaBerg: **Inconsistent handling of empty `comment_issue` forms.**

As written: `comment_issue: ""` is a validation error, `comment_issue: {body: ""}` is a validation error, but `comment_issue: {}` is a silent no-op. The first two are the strict reading, the third is the lenient one, and an author who writes `{}` gets no feedback that the action does nothing.

I'd make `{}` an error too, for one rule ("if the key is present, `body` must be non-empty"), which also lets `commentIssueRawPresent` be set unconditionally and simplifies the `Validate` loop.

AnaBerg (commented):
Review of `comment_issue`. Build and `go vet` are clean; `pkg/hub/pipeline`, `pkg/types/convert` and the new `comment_issue` tests pass. The 3 failures in `pkg/hub` (`TestDoneSignal*`) also fail on the merge base `6e533160`, so they are pre-existing and unrelated to this PR.

Test coverage here is genuinely good (~630 lines covering all four trackers, `move_issue` coexistence and ordering, `continue_on_error`, and the terminal-stage swallow), and the scalar/mapping parsing mirrors the existing `move_issue` shape.

Two findings I would treat as blockers before merge: the `continue_on_error` default killing the run, and the `fields: ["*all"]` polling change. The rest are follow-up material. Details inline.

AnaBerg (approved):

PR #637, opened 2026-08-18
chore(hack): add GitHub App private key rotation script

AnaBerg:
## What

Adds `hack/rotate-github-app-key.py`, a one-shot script to rotate the GitHub App private key used by an ElasticClaw hub.

There is no GitHub PAT in the Faster factory — git and `gh` inside claws authenticate with short-lived installation tokens minted from a GitHub App. So "changing the token" is really rotating that App's private key, which lives in **two** places that must stay in sync:

- `hub.yaml` → `github[]`
- `workspaces/<ws>/.elasticclaw-managed/github_apps.yaml`

## How

Everything goes through the hub API, so no service restart is needed (`PATCH /api/settings` persists and swaps the in-memory config; token providers are built per use).

1. Validates the PEM locally with `openssl rsa -check`
2. Reads current apps from `GET /api/settings` and `GET /api/workspaces/<ws>/github-apps`
3. Tests the new key against GitHub via `POST /api/settings/github/test` and **aborts before any write** if the key is bad or the installation is missing a required permission
4. Snapshots both remote files as `.pre-token-update.<ts>` over SSH (same convention as the 2026-07-02 rotation)
5. Writes hub.yaml (resending the full app list, since the patch replaces it) and the workspace app
6. Re-reads both and fails loudly, pointing at the backup, if the key or permissions look wrong afterwards

## Usage

```bash
hack/rotate-github-app-key.py --pem ~/Downloads/new-key.pem --check-only
hack/rotate-github-app-key.py --pem ~/Downloads/new-key.pem
```

Hub URL and admin token come from the `faster` profile in `~/.elasticclaw/config.yaml`.

## Testing

Read-only paths were exercised against the live Faster hub (profile parsing, both GETs, and the test endpoint with a throwaway key — correctly refused with `new key rejected: github get app 401` without writing anything). The write path has not been run; it will be exercised on the next real rotation.


AnaBerg:
Wrong premise — the factory's agents authenticate with a PAT that was configured on the server, not with the GitHub App key this script rotates. Closing while I locate where that PAT actually lives.

PR #677, opened 2026-09-03
refactor(web): give infrastructure alerts the notifier screen's own vocabulary

AnaBerg:
Follow-up to #675, which noted this as a known gap: the settings section that PR added
worked, but spoke a different control language than the screen it lives on. Lifecycle,
Channels and Scheduled reports are row cards plus a Dialog; Infrastructure alerts was a
native `select` and a grid of native checkboxes, inline, saving on every keystroke.

Stacked on `feature/infra-downtime-and-limit-notifications` so #675 stays reviewable.

## What changed

Each route is now a row card in the pattern the screen already uses: the channel as a
mono name, the "All alerts" / "N of 6 alert types" badge in the Channels section's own
size and treatment, a line saying what it receives and where it posts, and Send test on
the card. Add and Edit open a Dialog with its own draft — the channel picker, an explicit
**All alert types** / **Only the types you pick** choice, the type checkboxes, Remove
bottom-left, and a Cancel that discards. Nothing reaches the hub until Save.

That radio pair replaces a guard rather than removing one. The hub reads `events: []` as
receive-all, so unchecking the last box used to be a silent flip from one alert type to
the firehose; the old section prevented it by disabling the last checked box and
explaining why in text. Making the choice explicit says the same thing in the shape of
the question instead of in a constraint, and an empty "Only these" list simply cannot be
saved.

## Behaviour that had to survive

Every guard in that section is scar tissue from a review round on #675, so the rework was
done against an inventory rather than from scratch: the master switch gating (a route
pointing at a deleted channel makes *every* save on the screen fail, channels and reports
included), the in-section "Last save failed" strip, the channel-delete pause notice, the
receive-all trap above, per-route accessible names, the picker omitting channels another
route already uses, the PATCH carrying `notifiers`/`lifecycle`/`scheduled` through
unchanged, and the clamp that turns alerts off with the last route.

## Review

Behaviour-parity review (Claude), independent state/save-contract review (gpt-5.6), and a
hands-on pass against a running hub. Four findings fixed: routes were addressed by array
index rather than by channel; the badges rendered a size smaller than the identical
strings directly above them; the orphan card's amber remedy was dimmed to 50% by the
disabled wrapper; and removing the last route switched alerts off without the section's
own pause notice ever appearing. One reviewer scenario was rejected as unreachable in the
current UI, with the fix kept as hardening — the reason is in the commit.

`npm run typecheck` clean; `npm run lint` unchanged at 11 pre-existing warnings.

### DIFF
--- a/web/app/settings/[[...parts]]/settings-content.tsx
+++ b/web/app/settings/[[...parts]]/settings-content.tsx
@@ -6772,13 +6772,21 @@ function InfraNotificationsSection({ settings, onSave, saving, pause }: { settin
   // the hub until Save.
   const [showDialog, setShowDialog] = useState(false)
   const [dialogMode, setDialogMode] = useState<"add" | "edit">("add")
-  // The ROW being edited. Routes have no id, so position is the only address.
-  const [editIndex, setEditIndex] = useState<number | null>(null)
+  // The ROUTE being edited, addressed by the channel it posted to when the
+  // dialog opened — null while adding. Identity, not position: the hub allows
+  // exactly one infrastructure route per channel, so `via` is a key, whereas an
+  // index is only a key for as long as the list holds still. It does not have
+  // to hold still. Every save on this page re-fetches the whole config, and the
+  // routes that come back are whatever the hub now holds — reordered, or short
+  // one entry another writer dropped. An index captured before that round trip
+  // would then address a DIFFERENT route, and the save would quietly overwrite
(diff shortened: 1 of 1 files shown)


AnaBerg:
Consolidated into #675 — the whole change ships as one PR. GitHub closed this as merged when its base branch took the commits (61160c31); nothing was dropped.

PR #679, opened 2026-09-03
Cleanup whitespace in Makefile

xavpaice:
Minimal change to trigger e2e tests to confirm the status of them.


xavpaice:
tests pass now, this PR doesn't actually add any value so closing.

PR #684, opened 2026-09-10
fix(hub): create workflow_v2_runs timeout index after adding the column

xavpaice:
Fixes a production migration failure when upgrading to the latest code.

## Problem

The idx_workflow_v2_runs_timeout index references the timeout_at column, but it was created inside the initial workflowv2.Migrate SQL block before addColumnIfMissing added timeout_at. On databases where workflow_v2_runs already existed without timeout_at, the migration failed with:

    SQL logic error: no such column: timeout_at (1)

## Fix

- Move the index creation to after addColumnIfMissing has added timeout_at.
- Add a regression test that creates a legacy workflow_v2_runs table without timeout_at, runs Migrate, and asserts the column and index are present.

## Verification

    go test ./pkg/hub/workflowv2 -run TestMigrateAddsTimeoutAtIndex -v
    go test ./pkg/hub/workflowv2 -count=1
    go vet ./pkg/hub/workflowv2

All pass.


xavpaice:
This is an urgent fix, folks who install 2026.9.10-beta.1 may find the service unable to start without this patch.  Merging without review.

Issue #526, opened 2026-07-21
Make workspace flake tools available to agents and deterministic workflow steps


marccampbell:
## Problem

Workspace files already support a root `flake.nix` and optional `flake.lock`, and `elasticclaw-config.yaml` can opt into Nix with `nix: true`. However, a workspace flake does not currently provide one reliable tool environment for both the agent and deterministic workflow `run` actions.

This matters for workspace-specific CI CLIs. For example, a workspace may use Depot while another uses a different CI system. The Depot CLI should be declared by that workspace's flake, pinned by its lock file, and be available to both the claw and a deterministic step such as:

```yaml
on_enter:
  run:
    command: depot ci ...
```

without baking Depot into every provider image or manually installing it in every workflow.

## What works today

- `pkg/config/hub.go:ReadTemplateFiles` includes workspace-root `flake.nix` and `flake.lock`, so `elasticclaw workspace push` can persist them.
- `nix: true` installs Nix on the claw.
- The generic claw-bridge bootstrap used by exe.dev/Replicated detects a staged workspace flake, copies it to `~/.elasticclaw/flake`, creates `~/.elasticclaw/flake-run`, and starts the OpenClaw gateway with `nix develop`.
- A deterministic command can sometimes work around the missing integration by explicitly initializing Nix and invoking `nix develop` itself.

For example, the current Daytona path may require workflow commands to repeat implementation details similar to:

```yaml
command: . /nix/var/nix/profiles/default/etc/profile.d/nix-daemon.sh && nix develop --accept-flake-config -c depot ci ...
```

That is not a good workspace contract and is not provider-portable.

## Gaps in the current implementation

1. Deterministic workflow commands bypass the flake environment.
   - `executePipelineCommand` runs a fresh provider command equivalent to `cd "$HOME/.openclaw/workspace" && <command>`.
   - It does not call the generated `~/.elasticclaw/flake-run` wrapper or otherwise enter `devShells.default`.
   - A gateway process running inside `nix develop` does not cause a separate Daytona/SSH/Lambda exec session to inherit that environment.

2. Daytona does not set up or enter the workspace flake.
   - `bootstrapDaytona` can install Nix when the database `nix` flag is set, but it does not evaluate the workspace flake, create the wrapper, or start the gateway inside `nix develop`.
   - The Daytona gateway is started before workspace files are written, so the workspace flake cannot affect the agent runtime.

3. Flake presence does not consistently imply Nix support.
[...]

PR #692, opened 2026-09-16
fix(workflow-v2): run exec effects in the staged workspace after bootstrap

### LINKED ISSUE #691: V2 workspace assembly does not deploy workspace files/scripts or repository checkouts
**Problem**

When a workspace uses `schema_version: v2`, the published sandbox does not contain the workspace definition files (`scripts/`, `AGENTS.md`, `*.md`, etc.) or the repositories configured in `repositories`. Only `flake.nix`, `flake.lock`, and `openclaw-workspace-state.json` are staged under `~/.openclaw/workspace`.

This breaks any v2 workflow that relies on workspace helper scripts or on repository checkouts, because the files are simply not present.

**Environment**

- Hub: `https://factory.repldev.com`
- Provider: `replicated`
- Workspace name: `replicated-troubleshoot`

**Observed behavior (v2)**

A diagnostic `exec.run` of `ls -la ~/.openclaw/workspace` shows only:

```
flake.lock
flake.nix
.git/
openclaw-workspace-state.json
```

A workflow `exec.run` that calls `bash scripts/prepare-deps-branch.sh ...` fails with:

```
bash: scripts/prepare-deps-branch.sh: No such file or directory
exit code 127
```

**Observed behavior (v1, for comparison)**

Pushing the same workspace directory with `schema_version: v1` produces a full sandbox:

```
~/.openclaw/workspace/scripts/summarize_local_checks.py
~/.openclaw/workspace/AGENTS.md
~/.openclaw/workspace/CONTEXT.md
~/.openclaw/workspace/elasticclaw-config.yaml
~/.openclaw/workspace/troubleshoot/AGENTS.md
~/.openclaw/workspace/troubleshoot/CLAUDE.md
~/.openclaw/workspace/troubleshoot/.factory-workflows/dependency-update.yaml
...
```

So the v1 assembly path correctly copies workspace files and checks out configured repositories; the v2 path does not.

**Expected behavior**

A v2 workspace push should stage the same contents as v1: the full workspace directory (scripts, instruction files, config) plus the repositories declared in `repositories`. This matches the v2 documentation which says workspace files and scripts are part of the published workspace.

**Reproduction**

1. Create a workspace directory with `schema_version: 2`, a `scripts/` directory, and a repository in `repositories`.
2. `elasticclaw workspace push <name> --path <dir>`
3. Trigger any v2 workflow that runs `exec.run: { command: "bash scripts/example.sh" }` or expects a repository checkout.
4. Inspect the run facts; the command exits 127 because `scripts/example.sh` is missing.

**Impact**

We cannot migrate our dependency-update workflows to v2 because they depend on shared workspace scripts and repository checkouts. We are currently blocked on merging our v2 conversion PR.

xavpaice:
Fixes #691

Three fixes, validated end-to-end on a live Replicated hub during reporting (see validation comment).

## 1. Run exec effects in the staged workspace after bootstrap

The #691 report ("v2 sandbox contains only flake files, no scripts/repos, exec exits 127") was real, but the v2 *assembly* data was correct — two defects made runs execute against an unstaged workspace:

- **Dispatch raced bootstrap.** The control websocket is authorized purely by the run attempt, so replicated/exedev bridges (started inside the bootstrap script) received `exec.run` assignments while the hub was still writing workspace files and cloning repositories. v1 starts workflows from `promoteBootstrapReadyClaw` only after `bootstrap_ok=1`; v2 had no gate. `sendWorkflowV2ControlOutbox` now holds delivery until `workflowV2ClawDispatchReady` (reuses `allowWakeBeforeBootstrap`; providers that stage before the bridge starts are immediate), with a repeated hold log (~30s) so a waiting run is visible.
- **exec.run / dependency.update had no working-directory contract.** The bridge ran `sh -c <command>` from `$HOME`, so `bash scripts/foo.sh` exited 127 even after staging. `bridgeWorkspaceCommand` mirrors v1's `buildWorkspaceRunCommand`: `cd "$HOME/.openclaw/workspace" &&` plus `~/.elasticclaw/flake-run bash -lc` wrapping for flake workspaces (#526 devShell contract).

## 2. Cancel bound runs when a claw stops terminally

Terminal claw transitions finished the v1 `workflow_runs` row but never touched `workflow_v2_runs` — `cancelWorkflowV2RunForClaw` was only called from the workflow-creator provisioning paths. Bootstrap failures from the VM sync loop, the hub-restart fail-closed path, watchdog stops, and reaper deletes left active v2 runs holding a dead claw; with the dispatch gate those runs sat silently until the 24h run reaper. `finishClawTerminalTx` now cancels any active/suspended v2 run bound to the claw after the terminal commit (no-op when none; the nested `maybeFinishWorkflowV2Parent` recursion terminates at the status guard).

## 3. Resolve dotted fact keys in transition guards

Found live on the test hub: `sbctl-update-go`'s `prepare` exec.run completed (exit 0, facts written, receipt accepted) but `prepare_ready` never fired and the run parked at `state_version 1` — because

```yaml
when:
  exec.last_run:      # YAML: one key literally named "exec.last_run"
    succeeded:
      equals: true
```

parses to a literal dotted key, while facts load as nested maps (`exec` → `last_run` → `succeeded`), so the guard could never match. Dotted keys are the canonical fact namespace everywhere else (`workflow_v2_facts.fact_key`, `include_facts` SQL lookups, `exec.last_run.*` docs) — the predicate evaluator was the outlier. It now resolves dotted predicate keys as paths into the nested fact tree (literal-key lookup preserved). Additionally, accepted-without-transition receipts now carry a reason ("N transition(s) for <kind> in state <state> listened but evaluated false") so this failure class never parks silently again.

## Tests

- `TestWorkflowV2ControlOutboxHeldUntilClawBootstrapCompletes` — hold → release on `bootstrap_ok=1`
- `TestRunExecCommandResolvesWorkspaceRelativePaths` — real `bash scripts/foo.sh` from the live workspace (the exit-127 repro)
- `TestBridgeWorkspaceCommand*` — command construction with/without flake
- `TestWorkflowV2TriggerStagesWorkspaceFilesAndRepositories` — v2 assembly parity with v1 (template_files + github_repos)
- `TestFinishClawTerminalTxCancelsBoundWorkflowV2Run` — terminal stop cancels the bound run
- `TestApplyCommandReceiptExecRunTransitionsOnDottedFactGuard` — the parked-run repro (fails before fix 3)
- `TestMatchingTransitionResolvesDottedGuardKeys`, `TestMatchPredicateDottedKeysResolveNestedFacts` — dotted guard resolution and listener diagnostics

`go test ./pkg/hub/ ./pkg/hub/workflowv2/ ./pkg/types/v2/ ./cmd/claw-bridge/ ./cmd/`, `go vet`, and `gofmt` pass.

## Notes

- `docs/v2-schema-reference.md` has a matching exec.run contract note (cwd, flake-run, bootstrap wait), but `docs/` is untracked in this repo, so it is not part of the PR.
- Follow-up filed as #693: upgrade pinned OpenClaw 2026.7.1-2 → 2026.9.4 (upstream session-lock fixes).
- Hardening candidate (not included, effectively unreachable via the trigger path since `createClawFromWorkflowWithOptions` re-validates the v2 doc and hard-fails): `loadExternalWorkspaceConfigV2` returns a repository-less shell on load-time validation failure with only a stderr warning.


xavpaice:
## Live-hub validation (replicated-factory, built from this branch)

The full chain from #691 was verified on a deployed Replicated hub with a real v2 workspace (`replicated-troubleshoot` / `sbctl-update-go`, 3 GitHub repos, flake workspace):

1. **Staging** — sandbox contains all workspace files, `scripts/` (7 scripts), and all three repo checkouts (`sbctl/`, `troubleshoot/`, `troubleshoot.sh/`)
2. **Dispatch gate** — hub log shows `holding outbox for claw … until workspace bootstrap completes` (05:26:20, re-logged 05:26:50), released at `bootstrap_ok=1` (05:26:53)
3. **exec.run** — assignment delivered 05:26:58, executed in the live workspace: exit 0, `prepare-deps-branch.sh` created the deps branch, `exec.last_run.*` facts written, receipt accepted
4. **Fix 3 was found in this deployment** — the run then parked in `prepare` because its dotted-key guard (`exec.last_run: succeeded: equals: true`) never matched the nested fact tree; reproduced locally with the exact workflow and fixed in 97d36fe2. After that deploy the workflow advances through its states.

Diagnostics that made this tractable, for future reference: the bridge's local CLI (`claw-bridge control status`) exposes the run snapshot (state/phase/version/current task), and the hub DB (`workflow_v2_runs` / `workflow_v2_effects` / `workflow_v2_control_outbox` / `workflow_v2_events`) shows exactly which link stalled.

marccampbell (approved):

